PR #9, opened 2022-01-07
serialize figure captions with working references

dylangrandmont:
Requires https://github.com/curvenote/prosemirror-docx/pull/8

### DIFF
--- a/src/export/word/schema.ts
+++ b/src/export/word/schema.ts
@@ -1,5 +1,5 @@
 import { oxaLinkToId } from '@curvenote/blocks';
-import { Nodes, ReferenceKind } from '@curvenote/schema';
+import { CaptionKind, Nodes, ReferenceKind } from '@curvenote/schema';
 import { formatDatetime } from '@curvenote/schema/dist/nodes/time';
 import { HeadingLevel, IParagraphOptions, IRunOptions } from 'docx';
 import { defaultMarks, NodeSerializer } from 'prosemirror-docx';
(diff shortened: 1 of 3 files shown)


src/export/word/schema.ts
```diff
@@ -116,9 +132,20 @@ export function getNodesAndMarks(styles?: Styles) {
       state.closeBlock(node, styles?.iframe);
     },
     figure(state, node) {
+      const { id, numbered } = node.attrs as Nodes.Figure.Attrs;
+      // TODO: localize this ID
```
dylangrandmont, line 136: Can we actually remove this `TODO` @rowanc1 ? What should I check to confirm whether we need to "localize" this or not? (The cross references appeared to be working)

src/export/word/schema.ts
```diff
@@ -41,6 +41,22 @@ const defaultStyles: Styles = {
...
+  switch (kind) {
+    case CaptionKind.fig:
+      return 'Figure';
+    case CaptionKind.table:
+      return 'Table';
+    case CaptionKind.code:
+      // This is a hack, I don't think word knows about other things!
+      return 'Figure';
+    case CaptionKind.eq:
+      return 'Figure';
```
rowanc1, line 54:
```suggestion
    case CaptionKind.code:
    case CaptionKind.eq:
      // This is a hack, I don't think word knows about other things!
      return 'Figure';
```

src/export/word/schema.ts
rowanc1: We will need an options structure that looks like this:
https://github.com/curvenote/curvenotejs/blob/main/src/export/utils/localizationOptions.ts

The imageSrc isn't needed, but the other ones will be. I think you can actually import these directly from `curvenote`.

rowanc1 (commented):
Looks good, I think the todo needs to stay. Minor suggestion for the switch statement.

PR #13, opened 2022-01-18
Support Word Export References

dylangrandmont:
* Allow reference type to be specified such that docx state can be generated for it


src/export/word/template.ts
```diff
@@ -38,6 +38,10 @@ export async function defaultTemplate(data: LoadedArticle): Promise<Document> {
     if (!state) return;
     docxState.renderContent(state.doc);
   });
+  Object.values(article.references).map(({ state }) => {
+    if (!state) return;
+    docxState.renderContent(state.doc);
+  });
```
dylangrandmont, line 44: Some remaining tasks:
- [ ] Understand if the article order here will be deterministic and correct
- [ ] Need to a title for references if they exist (similar to `createArticleTitle` above?)
- [ ] Should we use `renderReference` from the [proposed prosemirror-docx](https://github.com/curvenote/prosemirror-docx/pull/9) changes? Probably doesn't matter here since there is no content between the main article content and the references (unlike custom templates)

src/export/word/template.ts
```diff
@@ -39,6 +34,15 @@ export async function defaultTemplate(data: LoadedArticle): Promise<Document> {
...
 
+  const references = Object.values(article.references);
+  if (references.length > 0) {
+    docxState.renderContent(createReferenceTitle());
+  }
+  references.forEach(({ state }) => {
+    if (!state) return;
+    docxState.renderContent(state.doc);
+  });
+
```
dylangrandmont, line 48: This should probably be a function but I didn't extract it to avoid having side effects on `docxState` be hidden in a function call

src/export/word/utils.ts
```diff
@@ -30,3 +30,12 @@ export function createSingleDocument(
   });
   return doc;
 }
+
+export function getDefaultSerializerOptions(buffers: Record<string, Buffer>) {
```
dylangrandmont, line 34: Util function we will make use of for private templates

rowanc1:
Looks good so far. Are you good to merge?

PR #27, opened 2022-02-01
Feat/import with references

stevejpurves:
Initial import script for markdown based content.

### DIFF
--- a/src/models.ts
+++ b/src/models.ts
@@ -87,11 +87,21 @@ export class Block extends BaseTransfer<BlockId, BlockDTO> {
 
   $fromDTO = blockFromDTO;
 
+  // TODO this isn't really the create url, its a get?
   $createUrl = () => `/blocks/${this.id.project}/${this.id.block}`;
 
   $recieve = blocks.actions.recieve;
 
   $selector = selectBlock;
+
+  static makeCreateUrl = (projectId: string) => `/blocks/${projectId}`;
+
+  static async create(session: ISession, projectId: string, data: JsonObject): Promise<BlockId> {
+    const { status, json } = await session.post(Block.makeCreateUrl(projectId), data);
+    if (status > 400) throw Error(`Could not create block, status: ${status}`);
+    if (!('id' in json)) throw Error(`Could not create block, invalid response: ${json}`);
+    return json.id as BlockId;
+  }
 }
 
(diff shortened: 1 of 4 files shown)


stevejpurves:
So currently this fails on any content blocks with a {cite:t}```id``` or {cite:p}```id``` in them.
The content I am importing as a test was generated by downloading this https://curvenote.com/@stevejpurves/geoscience/phase-and-the-hilbert-transform

So in this case our markdown does not roundtrip see https://github.com/curvenote/curvenotejs/pull/27/files#diff-85ddefb32fd9e0f743dd3c5c894331fe31359897fcdbd4ceaa597a5256a07202R41 for how that content is uploaded.

rowanc1:
This is the type of thing @fwkoch is working on now.

src/models.ts
```diff
@@ -87,11 +87,22 @@ export class Block extends BaseTransfer<BlockId, BlockDTO> {
...
   $fromDTO = blockFromDTO;
 
+  // TODO this isn't really the create url, its a get?
   $createUrl = () => `/blocks/${this.id.project}/${this.id.block}`;
 
   $recieve = blocks.actions.recieve;
 
   $selector = selectBlock;
+
+  static makeCreateUrl = (projectId: string) => `/blocks/${projectId}`;
```
stevejpurves, line 97: current `$createUrl` is not really the create URL!
maybe some renaming needed?

src/models.ts
```diff
@@ -87,11 +87,22 @@ export class Block extends BaseTransfer<BlockId, BlockDTO> {
...
+
+  static async create(session: ISession, projectId: string, data: JsonObject): Promise<BlockId> {
+    const { status, json } = await session.post(Block.makeCreateUrl(projectId), data);
+    if (status > 400) throw Error(`Could not create block, status: ${status}`);
+    if (!('id' in json))
+      throw Error(`Could not create block, invalid response: ${JSON.stringify(json)}`);
+    return json.id as BlockId;
+  }
```
stevejpurves, line 105: a static create function is probably the way to go but going to be common across more types.
Move most of this to base?
Also should we return the json package rather than just the id?

stevejpurves:
not really suggesting thas comes in as-is but more for discussion on how to approach the post/patch at model level and best design for the `import` cli

PR #28, opened 2022-02-01
added base array

stevejpurves:


stevejpurves:
tested locally using https://curvenote.com/@curvenote/testing/with-html-outputs

PR #32, opened 2022-02-02
⏬  fetch full html content when needed

### LINKED ISSUE #31: Only tiny pandas tables will be added to exports
Curvenote store large pandas tables in files, they are not immediately available over the API, instead, a download link is provided. The cli should be updated to download the full table (HTML) content when required.

stevejpurves:
Previously we were only rendering `output.contents` (which is truncated for large HTML strings) rather than fetching the full HTML payload from storage. fixes #31

### DIFF
--- a/src/export/utils/walkArticle.ts
+++ b/src/export/utils/walkArticle.ts
@@ -11,7 +11,6 @@ import {
 } from '@curvenote/blocks';
 import { DEFAULT_IMAGE_WIDTH, nodeNames, Nodes, ReferenceKind } from '@curvenote/schema';
 import { encode } from 'html-entities';
-import { EditorState } from 'prosemirror-state';
 import { getEditorState, getEditorStateFromHTML } from '../../actions/utils';
 import { Block, Version } from '../../models';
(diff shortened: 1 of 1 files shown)


src/export/utils/walkArticle.ts
```diff
@@ -59,14 +58,19 @@ function getFigureHTML(
...
+  const htmlOutput = version.data.outputs.find(
+    (output) => output.kind === OutputSummaryKind.html && Boolean(output.content),
+  );
+  if (!htmlOutput) return null;
+  let { content } = htmlOutput;
+  if (htmlOutput.link) {
+    const response = await fetch(htmlOutput.link);
+    if (!response.ok) return null;
+    content = await response.text();
+  }
```
stevejpurves, line 72: @dylangrandmont this was a gotcha with our HTML outputs that might be relevant to what you are doing. If the `content` of an output gets over a certain size, we truncate it and stach the whole thing on storage... so if a `link` field is present you'll need to fetch the html

dylangrandmont: Thanks for flagging, I definitely have encountered this

PR #33, opened 2022-02-04
fix/failed fetch

stevejpurves:
- fixed default template name
- added a rate limiter for fetching child blocks/versions and for fetching references in walkArticle, without any rate limiting the async map call could potentially launch 100s of concurrent requests if there are that many items in the export target.


stevejpurves:
Going to wait and see if our user comes online, in order to get access to test before committing to deploying this

stevejpurves:
tested locally - we now log an error for these references but continue

PR #37, opened 2022-02-08
🤐 Make latex logging quieter for CLI

rowanc1:


src/export/pdf/index.ts
```diff
@@ -32,9 +33,11 @@ export async function articleToPdf(
   });
 
   const buildPath = path.join(outputPath, '_build');
-  const CMD = `cd ${buildPath};latexmk -f -xelatex -synctex=1 -interaction=nonstopmode -file-line-error -latexoption="-shell-escape" ${tex_filename}`;
+  const CMD = `cd ${buildPath};latexmk -f -xelatex -synctex=1 -interaction=batchmode -file-line-error -latexoption="-shell-escape" ${tex_filename} &> ${tex_log_filename}`;
```
rowanc1, line 36: @stevejpurves this is the main change. Putting this to a file.

The main.log is still made, and I don't think the console output is different, but it is there if we want it.

src/export/tex/template.ts
```diff
@@ -19,17 +19,16 @@ export async function fetchTemplateTaggedBlocks(
...
-      template = await new Template(session, `tex/${opts.template}`).get();
-    } catch (err) {
-      // TODO - remove after private templates deploy settles
-      template = await new Template(session, opts.template).get();
+    session.log.debug(`Template: Fetching spec for '${opts.template}'`);
+    let requestedTemplate = opts.template.replace(/^tex\//g, '');
+    if (requestedTemplate.indexOf('/') === -1) {
+      requestedTemplate = `public/${requestedTemplate}`;
+      session.log.debug(`Template: Changing from '${opts.template}' to '${requestedTemplate}'`);
     }
```
rowanc1, line 27: I have also added public by default if it isn't added. This should also catch the /tex/ at the start.

rowanc1 (commented):
A few comments. if these are good with you @stevejpurves, can you merge!

src/export/pdf/index.ts
rowanc1: Yeah, does everrything nonstop does but doesn't print as many logs.

PR #40, opened 2022-02-22
feat/multi article

stevejpurves:
Building out a multi-article export


stevejpurves:
@rowanc1 rebased this and tested locally - this is ok to bring in

src/export/utils/walkArticle.ts
```diff
@@ -19,25 +19,25 @@ import { getImageSrc } from './getImageSrc';
 import { basekey } from './basekey';
 import { ISession } from '../../session/types';
 
-interface ArticleStateChild {
+export interface ArticleStateChild {
```
rowanc1, line 22: @stevejpurves this was intentional to hide these. And force us to look through `ArticleState` as the source of truth. Not too fussed though.

rowanc1 (commented):
Looks good.

PR #45, opened 2022-03-21
Feat/ejp multi article

stevejpurves:
Changes needed for finalise book export


stevejpurves:
ok bringing this in as changes are minimal and there are no collisions with the more recent `web`stuff

PR #58, opened 2022-04-15
Add linkBlock transformation and title/desc/thumb to link cache

fwkoch:
- Add linkBlock transformation similar to links.
- This extends the cache to hold title/desc/thumbnail in addition to url for internal links.

Todo:
- [ ] Actually cache thumbnails from the article


rowanc1:
Nice - I think this is probably just good to go!

Issue #56, opened 2022-04-08
Notebooks - follow links in markdown cells for local images


stevejpurves:
Currently during web builds, we do nothing to find and process local images that may be referenced in a markdown cell in a jupyter notebook.

We should start doing that when:
* the image included in available on a url
* the image included is available to us on disk relative to the notebook

Also, images can be referenced:
* in markdown syntax `![curvenote logo](./curvenote.png)`
* using an HTML `<img />` tag - (see also https://github.com/curvenote/curvenotejs/issues/55)

All these scenarios should be supported.

Currently, it seems in some cases that the build process correctly constructs links to local images when specified in markdown but the images are never made available in the `public` folder. e.g. https://transform.softwareunderground.org/2021-practical-python/t21-practical-seismic-with-python

**Note:** currently when we add notebooks to curvenote.com, images are lost :( we need more work on version control to fix that, so for this issue we have to manually make available and *locally referenced image* in order to test.

rowanc1:
🚀

PR #60, opened 2022-04-21
🌄 Improve image handling

### LINKED ISSUE #55: Notebooks render html in markdown cells (safely with limited schema?)
Currently in web builds we do nothing to handle html-in-markdown properly for notebooks, meaning that we end up rendering HTML source as plain text content.

It is common for notebook authors to use HTML in markdown cells to achieve things like this:

image

example notebook: https://github.com/gimli-org/transform2021/blob/main/1_Creating_a_subsurface_model.ipynb

Which currently renders in curvenote (and curvespace) as:

image

## Objective

Modify notebook processing during the web build process to handle HTML in markdown cells. 

The `processNotebook` function is where the specialised processing for notebooks happens at the moment. https://github.com/curvenote/curvenotejs/blob/feat/web/src/web/cache.ts#L46

When content/markdown cells are encountered we should process these (using schema/mdast) to sanitize (<script> removal) render the HTML appropriately.

**Note:** rendering images included in HTML, both local and via URL should be handled here 👉🏻 https://github.com/curvenote/curvenotejs/issues/56

fwkoch:
This works to address #56 

- [x] HTML images need to be rendered
  - **NOTE:** this was done by transforming all `html` nodes to `mdast`, so it will have wider-reaching effects than images. This may be a good thing (i.e. work to resolve #55) or a bad thing (cause unsafe, unwanted behavior).
- [x] Images from URLs render
  - These are now downloaded and cached when running `curvenote start`. Another valid option would be to leave the original URL and not overwrite with a CDN url.
- [x] Images from oxa links render
  - These are now correctly downloaded/cached/referenced
- [x] Local images no longer need to be moved to `images/` folder, local paths are respected

Takes us from:
image

to:
image

### DIFF
--- a/src/web/transforms/images.ts
+++ b/src/web/transforms/images.ts
@@ -1,19 +1,83 @@
+import { oxaLinkToId, VersionId } from '@curvenote/blocks';
 import { GenericNode, selectAll } from 'mystjs';
+import fetch from 'node-fetch';
+import fs from 'fs';
 import path from 'path';
+import mime from 'mime-types';
 
-import { Root } from './types';
+import { Root, TransformState } from './types';
+import { Options } from '../types';
 
-export function transformImages(mdast: Root) {
+export function serverPath(opts: Options) {
+  const buildPath = opts.buildPath || '_build';
+  return `${buildPath}/web`;
+}
+
+export function publicPath(opts: Options) {
+  return path.join(serverPath(opts), 'public');
+}
+
+export function imagePath(opts: Options) {
+  return path.join(publicPath(opts), '_static');
+}
+
+export async function transformImages(mdast: Root, state: TransformState) {
   const images = selectAll('image', mdast) as GenericNode[];
-  images.forEach((image) => {
-    // leave urls alone
-    if (
-      image.url.toLowerCase().startsWith('http:') ||
-      image.url.toLowerCase().startsWith('https:')
-    ) {
-      return;
-    }
-    // assumes all images referenced in myst content are moved to public/_static
-    image.url = `/_static/${path.basename(image.url)}`;
-  });
+  await Promise.all(
+    images.map(async (image) => {
+      const oxa = oxaLinkToId(image.url);
+      const isUrl =
+        image.url.toLowerCase().startsWith('http:') || image.url.toLowerCase().startsWith('https:');
+      // Leave non-oxa URLs alone
+      if (!oxa && isUrl) return;
+      // Assume remaining non-oxa paths are local images
+      if (!oxa) {
+        image.url = `/_static/${path.basename(image.url)}`;
+        return;
+      }
+      // Otherwise, fetch oxa image and save locally
+      const versionId = oxa?.block as VersionId;
+      if (!versionId?.version) return;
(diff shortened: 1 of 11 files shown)


src/web/transforms/images.ts
```diff
@@ -1,12 +1,98 @@
...
+  await Promise.all(
+    images.map(async (image) => {
+      const session = state.cache.session;
+      const oxa = oxaLinkToId(image.url);
+      let file: string;
+      if (oxa) {
+        // If oxa, get the download url
+        const versionId = oxa?.block as VersionId;
+        if (!versionId?.version) return;
+        const url = `/blocks/${versionId.project}/${versionId.block}/versions/${versionId.version}`;
```
stevejpurves, line 44: maybe good to have this as a `versionIdToURL(id: versionId)` somewhere - looks like we do this a couple times but haven't dry'ed it up

src/web/transforms/images.ts
```diff
@@ -1,12 +1,98 @@
...
+        }
+        file = await downloadAndSave(downloadUrl, `${versionId.block}.${versionId.version}`, state);
+      } else if (isUrl(image.url)) {
+        // If not oxa, download the URL directly and save it to a file with a hashed name
+        file = await downloadAndSave(image.url, hashUrl(image.url), state);
+      } else {
+        // Assume non-oxa, non-url paths are local images relative to the config.section.path
+        const fullPath = path.join(state.folder, image.url);
+        if (!fs.existsSync(fullPath)) {
+          console.log(fs.readdirSync('./'));
```
stevejpurves, line 60: stray console.log

src/web/transforms/images.ts
```diff
@@ -1,12 +1,98 @@
...
+  return url.toLowerCase().startsWith('http:') || url.toLowerCase().startsWith('https:');
+}
+
+export async function transformImages(mdast: Root, state: TransformState) {
   const images = selectAll('image', mdast) as GenericNode[];
-  images.forEach((image) => {
-    // assumes all images referenced in myst content are moved to public/_static
-    image.url = `/_static/${path.basename(image.url)}`;
-  });
+  await Promise.all(
```
stevejpurves, line 35: not awaiting here and returning this promise is also an option *if* you can await on `transformImages` ok... which it look like you are?

src/web/transforms/index.ts
```diff
@@ -63,25 +64,31 @@ export async function transformMdast(
...
     transformKeys,
   ].forEach((transformer) => {
-    transformer(mdast, { references, citeRenderer, cache, frontmatter });
+    transformer(mdast, { references, citeRenderer, cache, frontmatter, folder });
     cache.session.log.debug(
       toc(`Processing: "${name}" - ${transformer.name.slice(9).toLowerCase()} in %s`),
     );
   });
+
+  await transformImages(mdast, { references, citeRenderer, cache, frontmatter, folder });
```
stevejpurves, line 89: this is the only place where `transformImages` is called? and we can await it

stevejpurves:
left some comments but minor, this looks good - just waiting to test.

stevejpurves (approved):

PR #64, opened 2022-04-23
Add transform to nest un-blocked md in a block

### LINKED ISSUE #62: raw markdown from github doesn't render nicely
If i take a markdown file from github e.g. https://github.com/Zabamund/T22_getting_started/blob/master/Installation_and_setup.md

and drop it into a content folder on disk (updating the `_toc.yml`)  then when I start curvenote and check the page the rendered markdown does not look anything like the original, and local images are missing.

image

image

fwkoch:
This resolves #62 

Rather than modifying the md with `+++` as suggested in the issue, this just passes the mdast through a transform that nests everything in a `block` node if no other `block` nodes are present.

image

Unrelated, but this also adds an html-to-mdast-transform handler for comments like this so they don't show up in curvespace:

image

(upstreamed to mystjs too: https://github.com/executablebooks/mystjs/commit/e9320df9068d8ade41e515c2f753c0348dd4b673 )

### DIFF
--- a/src/web/transforms/blocks.ts
+++ b/src/web/transforms/blocks.ts
@@ -1,0 +1,9 @@
+import { GenericNode, select } from 'mystjs';
+import { Root } from './types';
+
+export function ensureBlockNesting(mdast: Root) {
+  if (!select('block', mdast)) {
+    const blockNode = { type: 'block', children: mdast.children as GenericNode[] };
+    (mdast as GenericNode).children = [blockNode];
+  }
+}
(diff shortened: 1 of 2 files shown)


stevejpurves:
Nice - ok that sounds like a good solution!
And yeah I saw that `html` comment and removed it from the content

src/web/transforms/blocks.ts
```diff
@@ -0,0 +1,9 @@
+import { GenericNode, select } from 'mystjs';
+import { Root } from './types';
+
+export function ensureBlockNesting(mdast: Root) {
+  if (!select('block', mdast)) {
+    const blockNode = { type: 'block', children: mdast.children as GenericNode[] };
+    (mdast as GenericNode).children = [blockNode];
+  }
+}
```
stevejpurves, line 9: looks good - so simple :)

Issue #76, opened 2022-04-25
Cannot Find Image


rowanc1:
Cannot find image, when it is a base64 image.

We should write this to the correct file.

![image](https://user-images.githubusercontent.com/913249/165146484-f2cfa1c0-de3d-41f2-bf02-11ebd04248e4.png)

stevejpurves:
not sure if this overlaps with https://github.com/curvenote/nbtx/issues/4

fwkoch:
nbtx handles base64-encoded jupyter outputs. However, this doesn't support other base64 strings in `image` MyST nodes, e.g. using `data:` URIs in markdown that looks like:
```
![](data:image/png;base64,iVBORw0KGgo...)
```

PR #89, opened 2022-04-28
🌠 Support inline base64 encoded images in markdown

fwkoch:
Resolves: #76


fwkoch:
image

rowanc1:
Looks solid. Haven't tested, but I think this is just good to go -- feel free to merge/squash it in. 🚀

Issue #88, opened 2022-04-28
Replace link with DOI that can be resolved with a citation


rowanc1:
We should pull citation information into a bibtex file on demand if we can and treat the link as a text citation. In Curvenote this should be clickable and bring up a tooltip!

![image](https://user-images.githubusercontent.com/913249/165672345-b5895fa7-7f68-4c56-8cf7-4df2dca4174f.png)

stevejpurves:
ok, so part of this is about curvenote import and roundtrip when working with notebooks, but another part could be also supporting these properly as citations in curvespace too, with popup etc... yeah?

rowanc1:
I think if there is a function that walks a raw notebook and hits our DOI service (e.g. openalex) and then pulls the content locally as a bibtex, then everything should work fine. A few additions needed to make the citations indexed by doi, but that should be straight forward.

PR #109, opened 2022-05-20
🪄  Transform links with DOIs into citations

rowanc1:
See #88


rowanc1:
This is motivated by @johnjarmitage's site as a test case!

All references in that example work.
image

And creates a references section at the bottom:
image

No changes to the text are necessary.

Issue #114, opened 2022-05-23
Add smoke tests


stevejpurves:
We need some minimal tests on CI covering different node versions, currently that is node/npm: 14/6, 16/8

PR #115, opened 2022-05-23
🔥 added a minimal smoke test that invokes the `cli`

stevejpurves:
The purpose is to get an invocation of the CLI into the tests to ensure that the install worked and a minimal cli `web build` command runs successfully. Made this as minimal as possible in light of structural changes coming in soon.

Addresses: #114


src/cli/services/web.ts
```diff
@@ -15,6 +15,12 @@ function makeCleanOption() {
 function makeForceOption() {
   return new Option('-f, --force', 'Remove the build directory and re-install').default(false);
 }
+function makeContentOnlyOption() {
+  return new Option(
+    '-o, --content-only',
+    'Process local content only, do not re-install the build folder if missing',
+  ).default(false);
+}
```
stevejpurves, line 23: Naming here could be improved maybe? cc @rowanc1

src/web/types.ts
```diff
@@ -44,5 +44,6 @@ export type Options = {
   clean?: boolean;
   force?: boolean;
   branch?: string;
+  contentOnly?: boolean;
```
stevejpurves, line 47: adds to global Options type :/

tests/smoke.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const exec = util.promisify(execWithCb);
+
+describe('CLI Smoke Tests', () => {
+  test('an example site', async () => {
+    try {
+      const { stdout } = await exec('curvenote web build -o', { cwd: 'tests/example' });
+      console.log(stdout);
+    } catch (error) {
+      expect(error).toBeNull();
```
stevejpurves, line 12: a better way to assert here.... 🤔

stevejpurves:
I tried manually reverting the npm-shrinkwrap here but after running `npm i` locally I appear to get the same changes -- using node `v16`

Issue #102, opened 2022-05-18
curvenote nodejs > 17


johnjarmitage:
Hi Rowen and Steve,
Just trying to install curvenote cli on my work computer (no sudo previlages). I used mamba to insall nodejs and by default it takes the latest version 17.8.0. When running `curvenote init` it fails, as it wants nodejs < 17.0.0. Is this the expected behaviour? I will downgrade nodejs to v16 and test it again.
Ciao,
John

johnjarmitage:
Install with nodejs 16.14.2 works fine.

stevejpurves:
As you found @johnjarmitage, `curvenote start` also has issues on 16. 14 is working though.

Usually nodejs kill odd numbered releases, as they are short lived. LTS and stable releases are on even numbers. So It's a bit weird that mamba gives you 17.

## TODO
- [ ] fix `curvenote start` on >= 16
- [ ] test v18
- [ ] add CI smoke test on the different versions; 14, 16 (,18)

rowanc1:
File watching is discussed in #112, and should now be fixed.

stevejpurves:
Regarding supporting >=17, we probably won't do this until v18 moves to "Active" status (see: https://nodejs.org/en/about/releases/), currently v17 and v18 are in current phase, where library support is being updated across the package ecosystem.

Opening a new issue to track CI smoke tests here: https://github.com/curvenote/curvenotejs/issues/114

PR #117, opened 2022-05-23
🚒 Add node engines that are supported

rowanc1:
See #1

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -97,6 +97,9 @@
         "ts-jest": "^27.1.4",
         "ts-node": "^10.7.0",
         "typescript": "^4.6.3"
+      },
+      "engines": {
+        "node": ">=14"
       }
     },
     "node_modules/@ampproject/remapping": {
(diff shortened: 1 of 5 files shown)


npm-shrinkwrap.json
```diff
@@ -97,6 +97,9 @@
         "ts-jest": "^27.1.4",
         "ts-node": "^10.7.0",
         "typescript": "^4.6.3"
+      },
+      "engines": {
+        "node": ">=14"
```
rowanc1, line 102: Followed remix package.json here.

src/cli/services/utils/utils.ts
```diff
@@ -19,7 +19,7 @@ ${chalk.bold('npm install -g npm@latest')}
 
 async function checkNodeVersion(session: ISession): Promise<boolean> {
   const checking = new Promise<boolean>((resolve) => {
-    check({ node: '>= 14.0.0 <17.0.0', npm: '>=6' }, (error, result) => {
+    check({ node: '>= 14.0.0', npm: '>=6' }, (error, result) => {
```
rowanc1, line 22: Removed 17 restriction, as it seems to work fine on 18, and added that to the testing.

rowanc1 (commented):
Added some notes!

src/cli/services/utils/utils.ts
rowanc1: See also #102

npm-shrinkwrap.json
stevejpurves: great, we didn;t have the enginers before -- good to add it 👍🏼

stevejpurves (approved):

PR #130, opened 2022-05-31
Frontmatter and Config validation

fwkoch:
This PR is a step towards https://github.com/curvenote/curvenotejs/issues/122 and https://github.com/curvenote/curvenotejs/issues/123.

- adds explicit `SiteFrontmatter`, `ProjectFrontmatter` and `PageFrontmatter` classes (rather than a single `Frontmatter` class)
- introduces object validation infrastructure with errors/warnings
- adds functions for validating all frontmatter classes
- implements clearer logic for resolving page/project frontmatter in accordance with the docs https://docs.curvenote.com/web/frontmatter

Still todo:
- This PR: make validation less strict. Log error messages but don't stop the build. (Maybe add a `strict` option to cli later that does make these stop the build)
- Next PR: Use this validation infrastructure to validate the remainder of the `curvenote.yml` config file on load.

TODO:
- [x] license is now licenses. Should it be? Should we accept both?
- [x] The project frontmatter is not loaded
- [x] url for the static asset is not necessarily a URL (it is a static path that must exist)
- [x] subject should be 40 char max


src/frontmatter/validators.ts
```diff
@@ -0,0 +1,393 @@
...
+  PageFrontmatter,
+  ProjectFrontmatter,
+  SiteFrontmatter,
+  Venue,
+} from './types';
+
+const CRT_CONTRIBUTOR_ROLES = [
+  'conceptualization',
+  'data curation',
+  'formal Analysis',
```
rowanc1, line 30: lowercase a?

fwkoch:
Frontmatter validation errors no longer stop build - they are logged to user and bypassed:

Screen Shot 2022-05-31 at 7 26 46 PM

stevejpurves:
getting some errors like:
```
```

stevejpurves:
I really like the clickable links to `yml` files I get on the console

stevejpurves:
bringing this in anyways, if there are more tweaks can stack them into the next PR

PR #140, opened 2022-06-02
enable client usage from the `launchpad`

stevejpurves:
* 👍  add `yes` option to `init`

### DIFF
--- a/src/sync/init.ts
+++ b/src/sync/init.ts
@@ -20,6 +20,7 @@ type Options = {
   branch?: string;
   force?: boolean;
   yes?: boolean;
+  domain?: string;
 };
 
 const WELCOME = async (session: ISession) => `
@@ -72,6 +73,7 @@ ${docLinks.overview}
  */
 export async function init(session: ISession, opts: Options) {
   if (!opts.yes) session.log.info(await WELCOME(session));
+  if (opts.domain) session.log.info(`Using custom domain ${opts.domain}`);
   let path = '.';
   // Initialize config - error if it exists
   if (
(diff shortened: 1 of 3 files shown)


rowanc1 (approved):
LGTM

src/sync/init.ts
```diff
@@ -136,8 +150,12 @@ export async function init(session: ISession, opts: Options) {
 
   session.log.info(await FINISHED(session));
```
fwkoch, line 151: Only show this if `!opts.yes` too, like the welcome message?

src/sync/init.ts
```diff
@@ -136,8 +150,12 @@ export async function init(session: ISession, opts: Options) {
...
 
-  const { start } = await inquirer.prompt([questions.start()]);
-  if (!start) {
+  let start = false;
+  if (!opts.yes) {
+    const promptStart = await inquirer.prompt([questions.start()]);
+    start = promptStart.start;
+  }
+  if (!start && !opts.yes) {
```
fwkoch, line 158:
```suggestion
  if (!start) {
```
I think this is all you need since `false` is the default, and it keeps `opts.yes` only applying to questions/output.

src/sync/init.ts
stevejpurves: it's completely silent then but yeah `-y` is for someone who knows what they are doing

src/sync/init.ts
stevejpurves: so "if we are not in `yes` mode, and user has decided not to `start`, send advisory message`

fwkoch: Ah I see 👍

fwkoch (approved):

PR #152, opened 2022-06-07
Improve page frontmatter

fwkoch:
- [x] Construct page frontmatter in the same way as project frontmatter
- [x] Replace notebook/article frontmatter with this function
- [x] Add `kind` to this frontmatter
- [x] Make sure it's tested!

### DIFF
--- a/src/export/markdown/index.ts
+++ b/src/export/markdown/index.ts
@@ -2,8 +2,14 @@ import path from 'path';
 import YAML from 'yaml';
 import { VersionId, KINDS, oxaLink, Blocks } from '@curvenote/blocks';
 import { createId, toMyst } from '@curvenote/schema';
-import { pageFrontmatterFromDTO } from '../../frontmatter/api';
-import { Block, Version } from '../../models';
+import { prepareToWrite } from '../../frontmatter';
+import {
+  pageFrontmatterFromDTO,
+  projectFrontmatterFromDTO,
+  saveAffiliations,
+} from '../../frontmatter/api';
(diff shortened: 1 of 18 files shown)


src/export/markdown/index.ts
```diff
@@ -22,40 +23,6 @@ type Options = {
...
-  return {
-    title: block.data.title,
-    description: block.data.description || '',
-    date: formatDate('date' in version.data ? version.data.date : version.data.date_created),
-    authors,
-    name: block.data.name,
-    oxa: oxaLink('', block.id),
-  };
-}
-
```
stevejpurves, line 58: ok, so this will have an immediate effect on export of `md`. I think we need to make the change to keep markdown export working as-is, by patching in the inherited frontmatter on export.

stevejpurves:
@fwkoch this looks good. I just re-pulled the content for the project where I had noticed the issue with authors being copied into the `ipynb` file and now on pull, we don't get the frontmatter copied. 👍🏼 

If I export `md` we get the frontmatter copied as expected
If I export `notebook` we do not get a frontmatter cell added, I think this is by design as `ipynb` files typically don't have them, it is inconsistent but maybe it is ok? cc @rowanc1 ??

fwkoch:
Regarding `notebook` export - yeah, there's an explicit `createFrontmatter` flag which is only set to `true` in `pull`. Very easy to change this behaviour if we want to make it consistent.

What about putting frontmatter in the notebook `metadata` instead of creating a separate markdown cell? is there some motivation for the separate cell that I'm missing?

Issue #87, opened 2022-04-28
Replace align, eqnarray environment with aligned*


rowanc1:
Currently katex fails at parsing environments, these should be translated to `\begin{aligned*}` and throw a warning in the console.

Look at:
* https://github.com/KaTeX/KaTeX/issues/1931#issuecomment-481259569
* https://github.com/microsoft/vscode/issues/119536

![image](https://user-images.githubusercontent.com/913249/165671625-d893f6c9-69ca-4160-98ab-146e22b60feb.png)

Test case:
```
\begin{eqnarray}
\mathbf{A}_{blocky} = 
\begin{bmatrix}
\mathbf{A}_{11} & \mathbf{A}_{12} \\
\mathbf{A}_{21} & \mathbf{A}_{22}
\end{bmatrix},
\end{eqnarray}
```

PR #161, opened 2022-06-13
🐛  KaTeX: Unknown column alignment: *

rowanc1:
image

Recovers for the input of `\begin{array}{*{16}c}` in this example (not supported in katex):
image

Also tests #87.

### DIFF
--- a/src/transforms/math.ts
+++ b/src/transforms/math.ts
@@ -1,3 +1,4 @@
+import chalk from 'chalk';
 import katex from 'katex';
 import { Math, InlineMath } from 'myst-spec';
 import { selectAll } from 'mystjs';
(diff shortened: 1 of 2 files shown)


src/transforms/math.ts
```diff
@@ -27,13 +28,25 @@ function replaceEqnarray(log: Logger, value: string, file: string) {
     .replace(/\\end{eqnarray}/g, '\\end{align*}');
 }
 
+type RenderResult = { html?: string; warnings?: string[]; error?: string };
+
+function removeWarnings(result: RenderResult, predicate: (warning: string) => boolean) {
+  if (!result.warnings) return result;
+  const nextWarnings = result.warnings.filter(predicate);
+  if (nextWarnings.length === 0) return { html: result.html };
```
fwkoch: This would lose `result.error` if it exists. Looks like we maybe never get that case, i.e. `RenderResult` is either `html/warnings` only or `error` only? But from this function that's not clear.

rowanc1: Nice, yeah, yours is better!

fwkoch (approved):
Don't have much context on the katex itself, but code seems sensible and makes sense.

PR #169, opened 2022-06-14
⬆️ Align tex export frontmatter with other frontmatter

fwkoch:
This PR moves towards using the correct frontmatter in exports and eliminating any duplicate logic around validation (or logic which is 95% duplicate with subtle differences...).

This should also move the `ExportConfig` to be more similar to `Project/SiteConfigs`...


src/export/validators.ts
```diff
@@ -0,0 +1,49 @@
...
+      },
+    );
+  }
+  return value;
+}
+
+/**
+ * Perform validation/corsion on jtex output
+ *
+ * TODO: these coersions could remain in memory and be passed directly to jtex, instead of written to file
```
rowanc1, line 32: Totally. Ideally jtex comes in to js land and we are writing less out to intermediate files.

rowanc1 (commented):
Took a quick look, @stevejpurves tested with @fwkoch.

Issue #141, opened 2022-06-02
Define `kind` on page


rowanc1:
Currently, we don't know whether a particular page is an article or a notebook, when we render them in curvespace.
So for example if we want to treat notebooks differently and render a Jupyter icon, or enable download, we can't adapt.

We should save a kind to the page. `kind=Article` or `kind=Notebook`. This should be on the top or the json.

fwkoch:
A few updates:
- Make sure this gets set on local pages
- Add it to mdast only, don't set as frontmatter on the page (not really something the user can change)

PR #181, opened 2022-06-14
📒 Move page kind from frontmatter to mdast cache

fwkoch:
`Kind` should not be on frontmatter - doesn't make sense for users to change this. Instead, it is just added to the `app/content` json, alongside `mdast`.

Should address #141 again.

### DIFF
--- a/src/store/local/actions.ts
+++ b/src/store/local/actions.ts
@@ -4,6 +4,7 @@ import fs from 'fs';
 import yaml from 'js-yaml';
 import { convertHtmlToMdast, GenericNode, selectAll } from 'mystjs';
 import { extname, join, dirname } from 'path';
+import { KINDS } from '@curvenote/blocks';
 import { SiteProject } from '../../config/types';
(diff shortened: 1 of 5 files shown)


src/store/local/actions.ts
```diff
@@ -223,7 +229,7 @@ export async function transformMdast(
       }),
     );
   }
-  const data: RendererData = { sha256, frontmatter, mdast, references };
+  const data: RendererData = { sha256, frontmatter, mdast, references, kind };
```
rowanc1, line 232:
```suggestion
  const data: RendererData = { kind, sha256, frontmatter, mdast, references };
```

rowanc1: Pretty sure order is maintained, and this should be at the top. :)

rowanc1 (commented):
Nice, yeah this looks like more like what I was expecting on the exported type. Will test quick and then merge.

rowanc1:
Love it.

image

image

PR #182, opened 2022-06-15
Enable jupyter outputs in web build articles

fwkoch:
This enables plotly, html, etc outputs in web builds when they are added to Articles on curvenote.com. It does not yet handle local (i.e. not versioned in curvenote) jupyter outputs updating when linked in curvenote articles.

Todo:
- [x] Move large output blobs out of site mdast and into public/_static (this is how they are handled already in Notebooks)

image

### DIFF
--- a/src/sync/pull.ts
+++ b/src/sync/pull.ts
@@ -35,8 +35,9 @@ export async function pullProject(session: ISession, path: string, opts?: { leve
   await projectToJupyterBook(session, project.id, {
     path,
     writeConfig: false,
-    createFrontmatter: true,
+    createNotebookFrontmatter: true,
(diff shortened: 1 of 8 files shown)


src/sync/pull.ts
```diff
@@ -35,8 +35,9 @@ export async function pullProject(session: ISession, path: string, opts?: { leve
   await projectToJupyterBook(session, project.id, {
     path,
     writeConfig: false,
-    createFrontmatter: true,
+    createNotebookFrontmatter: true,
     titleOnlyInFrontmatter: true,
+    keepOutputs: true,
```
stevejpurves, line 40: @fwkoch when do you think that we would have this set to false?

PR #189, opened 2022-06-16
🔗 Transform relative links to resolve correctly

fwkoch:
- [x] Resolve relative page link blocks to correct site paths
- [x] Resolve relative file link blocks to static content

Addresses: https://github.com/curvenote/curvenotejs/issues/176

### DIFF
--- a/src/store/local/actions.ts
+++ b/src/store/local/actions.ts
@@ -104,12 +104,12 @@ export async function loadFile(session: ISession, path: string) {
     switch (ext) {
       case '.md': {
         const mdast = parseMyst(content);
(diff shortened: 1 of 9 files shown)


fwkoch:
I can recreate those types of errors if I remove the `afterEach(() => mock.restore());` at the top of the test file. It's saying the `jest` stuff doesn't exist because it's still looking at the mock file system. However, I've never seen it with the `restore` statement present. 🤔

rowanc1:
Weird. And it is only in the actions.spec.ts, I got nothing...

PR #190, opened 2022-06-17
Add Site Frontmatter fallbacks on web build

fwkoch:
For now, the only field that is used from site frontmatter is `venue`. Also, this only affects web builds since exports do not have a "site" associated with them.

image


rowanc1:
Tested locally!

PR #197, opened 2022-06-17
🔖 fix spurious copy image errors

### LINKED ISSUE #196: bookmark links cause errors
image

these are caused by links such as:

`<a id="1.2"></a>`

and 
`[IP Simulation](#IP-simulation)`

stevejpurves:
fixes #196 

the links themselves then appear to work in the same way as before


stevejpurves:
@fwkoch this fixed the immeate error, but if you look at what existing behaviour was here, the local anchors didn't work - they force a new tab to open - see the table of contents near the top of the notebook here https://transform.softwareunderground.org/2022-what-is-mcmc/transform22-what-is-mcmc-solved it's the one we were looking at

fwkoch:
Yeah, makes sense, need to transform slightly still to add "internal" flag.

PR #204, opened 2022-06-20
Looking at changes from JB

rowanc1:


fwkoch (approved):
This all looks good to me - reducing thrown errors and better citation support 👍

PR #210, opened 2022-06-21
Add `kernelspec` and `jupytext` to page frontmatter

### LINKED ISSUE #162: Recognize `kernelspec` and `jupytext` frontmatter
If these are present, the markdown file is actually of kind notebook (see #141)

image

There should also be a pass to set any code-cell directives without langauge defined to the kernelspec.language.

This property is also sometimes nested under jupyter.kernelspec (which can be a transparent fallback).

fwkoch:
- [x] Support `kernelspec` in page fm with display_name, name, language, env, argv
- [x] Allow kernelspec to be defined on `jupyter.kernelspec`
- [x] Support `jupytext` in page fm with formats and text_representation
- [x] Transform code cells to include `kernelspec.language` if applicable

Resolves #162


rowanc1:
This looks good from a code review. Can chat and maybe see a test?

src/store/local/actions.ts
```diff
@@ -237,7 +238,14 @@ export async function transformMdast(
       }),
     );
   }
-  const data: RendererData = { kind, file, sha256, frontmatter, mdast, references };
+  const data: RendererData = {
+    kind: frontmatter.kernelspec || frontmatter.jupytext ? KINDS.Notebook : kind,
```
fwkoch, line 242: Is this the correct logic for setting `kind: Notebook` on jupytext md file?

rowanc1: This seems correct?! I will take this for a spin today.

PR #212, opened 2022-06-21
Enable init if project config is present

### LINKED ISSUE #211: `curvenote init`  should still init a site, if `curvenote.yml` exists and has no site section
If we have a local folder that already contains a `curvenote.yml` with a `project` section but with no `site` section, `curvenote init` should still be able initialize a site -- and honor the existing project information.

An additional use case for his would-be launchpad, where a user could add a `curvenote.yml` file to their github repo in order to control, frontmatter for the root project, including adding licenses etc...

fwkoch:
If project config exists locally, you may still `curvenote init` and it will be used. Note - if project config is present, nothing will be pulled; it will only build from local content. If you need to pull remote content, either do that explicitly first or `curvenote init` without the project config.

Resolves #211

### DIFF
--- a/src/licenses/licenses.spec.ts
+++ b/src/licenses/licenses.spec.ts
@@ -24,8 +24,13 @@ describe('validateLicense', () => {
   it('valid string coerces', async () => {
     expect(validateLicense('CC-BY-4.0', opts)).toEqual(TEST_LICENSE);
   });
-  it('valid license output errors', async () => {
-    expect(validateLicense(TEST_LICENSE, opts)).toEqual(undefined);
+  it('valid license object passes', async () => {
+    expect(validateLicense(TEST_LICENSE, opts)).toEqual(TEST_LICENSE);
+  });
+  it('invalid license object fails', async () => {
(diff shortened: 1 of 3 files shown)


stevejpurves:
Testing this one locally @fwkoch with this `curvenote.yml`

```
version: 1
project:
  license:
    content: CC-BY-4.0
    code: MIT
  open_access: true
```

and I hit errors

image

It allowed me to continue and serve the sie but had wiped out the license info (and maybe other frontmatter if it was there?):
```
project:
  license: {}
  open_access: true
  title: stevejpurves
```

fwkoch:
Thanks for catching this - `license` is a weird case where validation requires a string for code/content (as you have above) but outputs objects (which errors if re-validated...). Probably a couple fixes/improvements to go into this...

fwkoch:
@stevejpurves - give this another look: `license` validation was not idempotent before - as described in my previous comment. Now, re-validating passes transparently.

During `init` with an existing project config, we validate the config twice. I don't think that's intrinsically problematic, and now it no longer errors / ignores the licenses.

PR #213, opened 2022-06-21
🚨 Throw error when site promotion fails

### LINKED ISSUE #170: Promotion errors should cause exceptions / non success exits

fwkoch:
Resolves #170


stevejpurves:
tried this out locally, it works well, error is thrown at the end and if multiple domain with a problem with one of them, the good one still deploys :)

PR #214, opened 2022-06-22
path and exec cwd changes

stevejpurves:


stevejpurves:
Cross tested on mac ✅

Issue #221, opened 2022-06-23
Write something to the nav on local folder init


rowanc1:
Right now it is hard to edit it / know what to do next.

![image](https://user-images.githubusercontent.com/913249/175220845-b1cbc66e-4307-4f07-80ea-a0231a2db8bb.png)

And no nav in the website!

Issue #220, opened 2022-06-23
Auto toc should not travers into other projects


rowanc1:
This has me in the main directory, and a content folder with a project in it. There is duplicate content (although it builds nice and fast / dedup-ed!).

![image](https://user-images.githubusercontent.com/913249/175217314-9efdefbb-55c9-43a9-abdb-62e5ac2d6853.png)

stevejpurves:
is this related to https://github.com/curvenote/curvenotejs/issues/215 ??

Issue #215, opened 2022-06-22
`curvenote init` should honor existing sub projects


stevejpurves:
In https://github.com/curvenote/curvenotejs/pull/212 support was added for honoring an existing root `curvenote.yml` file with a `project` section and no `site` section, allowing the site to be initialized and the `project` data to be preserved.

However, projects in subfolders are ignored, for example if I have the following structure:

```
myrepo/
    curvenote.yml
    index.md
    example.ipynb
    content/
        project1/
            curvenote.yml
            notebook1.ipynb
        project2/
            curvenote.yml
            notebook2.ipynb
```

Currently running `curvenote init` produces a site with a single project, where here we would expect project boundaries to be picked up and for 3 projects to be listed in `site.projects` and potentially respective default entries made in the `nav` array?

PR #226, opened 2022-06-24
Improve init / toc generation with existing project config

fwkoch:
This PR addresses:
- [x] #220 - If a `curvenote.yml` exists in a subdirectory, the auto toc traversal will stop and not add that content to the project in the parent directory during `curvenote start`
- [x] #215 - If project configs exist on subfolders, these will be picked up during `curvenote init` and added to the `site.projects`. Users can then choose if they also want to create a root-level project config or not.
- [x] #221 - During `curvenote init` the `site.nav` is now populated with links to all projects found on `site.projects`

### DIFF
--- a/src/sync/init.ts
+++ b/src/sync/init.ts
@@ -106,15 +106,19 @@ export async function init(session: ISession, opts: Options) {
   let pullComplete = false;
   let title = projectConfig?.title || siteConfig.title;
   if (content === 'folder') {
-    if (!opts.yes) {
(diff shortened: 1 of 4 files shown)


stevejpurves:
@fwkoch yes this now hits the mark.
When I have content in the root directory, it doesn't add a "found" message like you mentioned at one point during our chat last week, but I think that is ok as it just works!

stevejpurves (approved):

Issue #208, opened 2022-06-21
If there is a _toc.yml in a folder, use it


rowanc1:
Right now there is a hard choice at the start to either use a toc of not, I think that we should be able to use a TOC if it exists in a folder, instead of reading all of the files in that directory.

stevejpurves:
this will also help with rendering existing jupyterbooks as expected, correct?

stevejpurves:
This works great when the `_toc.yml` is in the root folder exactly how I expect. TO Test I did a juptyerbook export, then modified the `_toc.yml` and `curvenote init`

stevejpurves:
in a slightly different test, I removed some of the `curvenote.yml` files from my blog and left it with  
a `_toc.yml` in a subfolder `content/blog` and made some edits there, to check that it would pick up on my `_toc.yml` file.

it did! but then it also generates this:

![image](https://user-images.githubusercontent.com/1473646/176010303-7a1a711c-e71b-4816-9f05-b9801455bc32.png)


I'm not sure if we are even thinking about covering this case? the only other reasonable behaviour I could see is that the root was honored, and the content/blog levels were not added *BUT* maybe that only makes sense in this case, where no other files were found in `content`

fwkoch:
Thanks for testing this @stevejpurves - For your second example, I think maybe we should modify it so the `Blog` level isn't there, i.e. the folder with the `_toc` in it. We know that TOCs will always have a single index file, which should just replace the folder.

However, I'm inclined to leave `Content` alone - seems like it's getting into lots of special cases. Lift up a single toc? Lift up multiple tocs? Do we do the same thing when auto-discovering files without toc? Easier to just let the user handle tidying up the structure or modifying the toc; should be pretty clear to them what's going on.

stevejpurves:
Yes, i agree trying to cover thos case would jntroduxe eother bad edge cases. Inlike you suggestion

PR #228, opened 2022-06-27
🚇 Respect _toc.yml page structure if present in project subfolder

fwkoch:
Addresses #208 

When auto-generating project contents, if a _toc.yml file is encountered, that structure is used instead of just adding all files.


stevejpurves:
just checked his but I think it still needs a small change to remove the Blog folder that you mentioned in the issue, correct @fwkoch ?

stevejpurves:
looks good!!

PR #240, opened 2022-07-05
🔖  Allow main as first page in TOC

rowanc1:


fwkoch (approved):
Looks good, this means `main.md` will be the _index_ file if `index.md` and `readme.md` do not exist. Also, if there are no markdown files, `main.ipynb` will be the index file.

PR #274, opened 2022-07-12
🎇 Thumbnails and tags on articles and notebooks

rowanc1:
Adds thumbnails and tags to articles that are downloaded.

### DIFF
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -83,8 +83,7 @@
     "prosemirror-model": "^1.18.1",
     "prosemirror-state": "^1.4.1",
     "redux": "^4.1.2",
-    "which": "^2.0.2",
-    "yaml": "^1.10.2"
+    "which": "^2.0.2"
   },
   "overrides": {
     "prosemirror-tables": "^1.1.1"
--- a/apps/cli/src/frontmatter/api.ts
+++ b/apps/cli/src/frontmatter/api.ts
@@ -1,4 +1,5 @@
 import { Block, oxaLink, Project } from '@curvenote/blocks';
+import { dirname, join } from 'path';
 import { affiliations } from '../store/local';
 import { selectAffiliation } from '../store/selectors';
 import { ISession } from '../session';
@@ -10,6 +11,9 @@ import {
   validatePageFrontmatterKeys,
   validateProjectFrontmatterKeys,
 } from './validators';
+import { downloadAndSaveImage } from '../transforms/images';
+
+const thumbnailsDirectory = 'thumbnails';
 
 export function saveAffiliations(session: ISession, project: Project) {
   session.store.dispatch(
@@ -51,6 +55,28 @@ export function projectFrontmatterFromDTO(
     count: {},
     ...opts,
   });
+}
+
+export async function pageFrontmatterFromDTOAndThumbnail(
+  session: ISession,
+  filename: string,
+  block: Block,
+  date?: string | Date,
+  opts?: Partial<Options>,
+): Promise<PageFrontmatter> {
+  const apiFrontmatter = pageFrontmatterFromDTO(session, block, date, opts);
+  if (block.links.thumbnail) {
+    const result = await downloadAndSaveImage(
+      session,
+      block.links.thumbnail,
+      block.name || block.id.block,
+      join(dirname(filename), thumbnailsDirectory),
+    );
+    if (result) {
+      apiFrontmatter.thumbnail = join(thumbnailsDirectory, result);
+    }
+  }
+  return apiFrontmatter;
 }
 
 export function pageFrontmatterFromDTO(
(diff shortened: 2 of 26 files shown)


apps/cli/package.json
```diff
@@ -83,8 +83,7 @@
     "prosemirror-model": "^1.18.1",
     "prosemirror-state": "^1.4.1",
     "redux": "^4.1.2",
-    "which": "^2.0.2",
-    "yaml": "^1.10.2"
```
fwkoch, line 87: 👍 Been needing to do this!

apps/cli/src/frontmatter/api.ts
```diff
@@ -10,6 +11,9 @@ import {
   validatePageFrontmatterKeys,
   validateProjectFrontmatterKeys,
 } from './validators';
+import { downloadAndSaveImage } from '../transforms/images';
+
+const thumbnailsDirectory = 'thumbnails';
```
fwkoch, line 16: `THUMBNAILS_DIRECTORY`? And maybe define in the same place we are defining `_build` and any other const-but-potentially-conflicting path names. Not a deal now.

apps/cli/src/frontmatter/api.ts
```diff
@@ -53,6 +57,28 @@ export function projectFrontmatterFromDTO(
   });
 }
 
+export async function pageFrontmatterFromDTOAndThumbnail(
```
fwkoch, line 60: I guess this is separate from `pageFrontmatterFromDTO` since we don't want thumbnail for tex.

apps/cli/src/store/local/actions.ts
```diff
@@ -251,6 +257,7 @@ export async function transformMdast(
     kind: frontmatter.kernelspec || frontmatter.jupytext ? KINDS.Notebook : kind,
     file,
     sha256,
+    slug: pageSlug,
```
fwkoch, line 260: Not sure where this is used - it feels slightly out of place; the source of truth for this should be the file info in the redux store. Duplicating it here means we need to keep those in sync.

apps/cli/src/store/local/selectors.ts
```diff
@@ -52,7 +53,6 @@ export function selectOxaLinkInformation(state: RootState, oxa: string) {
     title: fileInfo.title,
     description: fileInfo.description,
     url: info.url,
-    // TODO: thumbnail
-    thumbnail: undefined,
+    thumbnail: fileInfo.thumbnail || undefined,
```
fwkoch, line 56: Does this not respect `fileInfo.thumbnail === null`?

apps/cli/src/transforms/images.ts
```diff
@@ -45,50 +52,105 @@ async function downloadAndSave(session: ISession, url: string, file: string): Pr
   return extension ? `${file}.${extension}` : file;
 }
 
-export async function transformImages(session: ISession, mdast: Root, filePath: string) {
+export async function saveImageInStaticFolder(
```
fwkoch, line 55: This refactor looks fine.

apps/cli/src/transforms/images.ts
```diff
@@ -45,50 +52,105 @@ async function downloadAndSave(session: ISession, url: string, file: string): Pr
...
+    if (!image) {
+      session.log.debug(`${file}#frontmatter.thumbnail is not set, and there are no images.`);
+      return;
+    }
+    session.log.debug(`${file}#frontmatter.thumbnail is being populated by the first image.`);
+    thumbnail = image.sourceUrl || image.url;
+  }
+  if (!thumbnail) return;
+  session.log.debug(`${file}#frontmatter.thumbnail Saving thumbnail in static folder.`);
+  const result = await saveImageInStaticFolder(session, thumbnail, dirname(file));
```
fwkoch, line 150: Ah, so on markdown export, we copy thumbnails to `thumbnails` folder and rewrite frontmatter, then on transform, we coppy to `_static` and rewrite frontmatter again. I think this all makes sense.

apps/cli/src/store/local/selectors.ts
fwkoch: Ah, I guess it's ok, since this is just getting oxa link information. At this point, previously `undefined` values have already been attempted to resolve. So now for the oxa link, we say: "if the thumbnail was not defined and couldn't be resolved _or_ if it was always null, there is no thumbnail"

fwkoch (commented):
Looks good and makes sense I think. Nothing needs to be actioned from this review.

apps/cli/package.json
rowanc1: Nice!

apps/cli/src/frontmatter/api.ts
rowanc1: Done in next PR!

apps/cli/src/frontmatter/api.ts
rowanc1: It is also an async function that copies images... which didn't feel right.

apps/cli/src/store/local/actions.ts
rowanc1: This is copying over to the mdast json output, which was useful to me in a usecase in building the curvenote blog.

image

PR #278, opened 2022-07-17
🎇 Webp optimized thumbnails

rowanc1:


stevejpurves:
are we moving to using `changesets` now? should we be doing that consistently on PRs across the team?

rowanc1:
Yeah, I think so. Gives us a changelog and helps with publishing. I updated the readme to give a quick pointer, it is basically `npx changeset` and then follow the CLI. Later the version and publish commands work very well (tags, publishing, lock-step packages).

PR #282, opened 2022-07-19
Tabs and MyST Demo support

rowanc1:
Supporting things that are needed for nice MyST documentation:
* Tabs
* MyST renderer demo component
* `include` directive

This supports most of the things that are required for the current myst-spec and mystjs documentation.

myst-demo
tabs-dark
image

cc @choldgraf - I am working my way back to the myst documentation and site.

### DIFF
--- a/apps/cli/src/transforms/include.ts
+++ b/apps/cli/src/transforms/include.ts
@@ -1,0 +1,26 @@
+import fs from 'fs';
+import type { GenericNode } from 'mystjs';
+import { parseMyst, Root } from '../myst';
+import { selectAll } from 'mystjs';
+import { join, dirname } from 'path';
+import { ISession } from '../session';
+
+/**
+ * This is the {include} directive, that loads from disk.
+ *
+ * RST documentation:
+ *  - https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment
+ */
+export function includeFilesDirective(session: ISession, filename: string, mdast: Root) {
+  const includeNodes = selectAll('include', mdast) as GenericNode[];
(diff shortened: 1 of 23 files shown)


apps/cli/src/transforms/include.ts
```diff
@@ -0,0 +1,26 @@
...
+export function includeFilesDirective(session: ISession, filename: string, mdast: Root) {
+  const includeNodes = selectAll('include', mdast) as GenericNode[];
+  const dir = dirname(filename);
+  includeNodes.forEach((node) => {
+    const file = join(dir, node.file);
+    if (!fs.existsSync(file)) {
+      session.log.error(`Include Directive: Could not find "${file}" in "${filename}"`);
+    }
+    const content = fs.readFileSync(file).toString();
+    const children = parseMyst(content).children as GenericNode[];
```
fwkoch, line 23: What about non md included files - not supported (do we validate?)? included as raw text?

rowanc1: Yeah, good point. There are some more options on the import from RST that we could support as well.

rowanc1: https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment

rowanc1: I am going to leave this to the future. :)

choldgraf:
Can I just note that I think `{include}` in the context of interactive documentation writing / reading is a really interesting pattern? Obsidian has a similar functionality and I find it quite useful: https://help.obsidian.md/How+to/Embed+files

PR #283, opened 2022-07-20
Improve the clone function

rowanc1:


apps/cli/src/sync/utils.ts
```diff
@@ -24,6 +24,25 @@ export function getDefaultSiteConfig(title?: string): SiteConfig {
...
 
+export async function getDefaultSiteConfigFromRemote(
+  session: ISession,
+  projectId: string,
+  siteProject: SiteProject,
+): Promise<SiteConfig> {
+  const project = await new Project(session, projectId).get();
+  const remoteSiteConfig = await new RemoteSiteConfig(session, project.id).get();
+  const siteConfig = getDefaultSiteConfig();
+  siteConfig.title = project.data.title;
```
fwkoch, line 35: Should eventually do `remoteSiteConfig.title || project.data.title` but this doesn't really matter yet since we aren't providing nice ways for users to update site info independently of proj.

apps/cli/src/toc/manifest.ts
```diff
@@ -76,8 +77,12 @@ function getLogoPaths(
     // Look in the local public path
     logoName = join('public', logoName);
   }
-  if (!fs.existsSync(logoName))
-    throw new Error(`Could not find logo at "${origLogoName}". See 'config.site.logo'`);
+  if (!fs.existsSync(logoName)) {
+    session.log[opts.silent ? 'debug' : 'error'](
```
fwkoch, line 81: Feels a little heavy-handed to have an option that sets the log level... can't we just always `error`...? Not a big deal though.

apps/cli/src/toc/manifest.ts
```diff
@@ -76,8 +77,12 @@ function getLogoPaths(
...
+      `Could not find logo at "${origLogoName}". See 'config.site.logo'`,
+    );
+    return null;
```
fwkoch, line 84: This is good though, nice to be permissive with obvious errors logged.

apps/cli/src/sync/utils.ts
```diff
@@ -24,6 +24,25 @@ export function getDefaultSiteConfig(title?: string): SiteConfig {
...
+  siteConfig.logoText = project.data.title;
+  if (remoteSiteConfig.data.domains) siteConfig.domains = remoteSiteConfig.data.domains;
+  // Add an entry to the nav if it doesn't exist (i.e. empty list is fine)
+  if (!remoteSiteConfig.data.nav) {
```
fwkoch, line 39: Should this have:
```
else {
  siteConfig.nav = remoteSiteConfig.data.nav
}
```

fwkoch: I guess this is the same thing as for `title` etc. We need to do a better job of respecting the remote site config once users have an easy way to update....

apps/cli/src/sync/utils.ts
```diff
@@ -24,6 +24,25 @@ export function getDefaultSiteConfig(title?: string): SiteConfig {
...
+    siteConfig.nav = [{ title: project.data.title || '', url: `/${siteProject.slug}` }];
+  }
+  siteConfig.projects = [...siteConfig.projects, siteProject];
```
fwkoch, line 42: Yeah, this is all just slightly uncomfortable. Here, `siteConfig.projects` will _always_ be `[]` since the `siteConfig` is directly from `getDefaultSiteConfig`. So it would be a bit clearer if we just do `[siteProject]`... But, like, whatever... it doesn't matter at this point; maybe it would feel better if we weren't building on `getDefaultSiteConfig`. Make that `getDefaultLocalSiteConfig`, and here we just provide a different "remote" default.

apps/cli/src/store/api/reducers.ts
```diff
@@ -44,6 +45,16 @@ export const projects = createSlice({
   },
 });
 
+export const siteconfigs = createSlice({
```
fwkoch, line 48: SiteConfig on the API side of the redux store 👍

PR #285, opened 2022-07-25
⚠️  Audit site during build process

rowanc1:
Introduce `strict` and `check-links` parameters for the build process that can stop the build.

image


apps/cli/src/cli/web.ts
```diff
@@ -60,6 +64,8 @@ function makeDeployCLI(program: Command) {
     .addOption(makeForceOption())
     .addOption(makeYesOption())
     .addOption(makeCIOption())
+    .addOption(makeStrictOption())
```
fwkoch, line 67: Should these have `true` defaults here?

apps/cli/src/transforms/images.ts
```diff
@@ -111,7 +117,12 @@ export async function saveImageInStaticFolder(
     await fileObject.writeBase64(urlSource);
     file = fileObject.id;
   } else {
-    session.log.error(`⚠️  Cannot find image "${urlSource}" in ${opts?.sourceFile || filePath}`);
+    const message = `Cannot find image "${urlSource}" in ${opts?.sourceFile || filePath}`;
+    if (opts?.sourceFile) {
```
fwkoch: Some of this `if/else` could be dealt with in `addWarningForFile`... maybe just `addWarning`, and if `file` isn't provided, just `log` (or even save the error as non-file-specific...). Could also add `debug` handling in there too (log, don't save)?

I guess it's a trade off between reducing lines of code and making the `addWarning` function a bit more bloated in scope.

apps/cli/src/store/build/actions.ts
```diff
@@ -0,0 +1,63 @@
+import { ISession } from '../../session/types';
```
fwkoch: This doesn't quite feel like the right place for these functions? We are putting a bunch of non-store functionality under `store` (I'm already not super excited about all the stuff in `store/local/` - `actions.ts`, `notebook.ts`).

If I was looking around for these, I'd expect `checkLink` to be defined in `transforms/links.ts` (the only place it is used), and `addWarningToFile` to be in `utils` (it's pretty generic and used everywhere, plus we are already importing `store` in `utils`).

apps/cli/src/store/local/actions.ts
```diff
@@ -105,44 +111,45 @@ export function combineProjectCitationRenderers(session: ISession, projectPath:
   cache.$citationRenderers[projectPath] = combineRenderers(cache, ...project.citations);
 }
 
-export async function loadFile(session: ISession, path: string) {
+export async function loadFile(session: ISession, file: string) {
```
fwkoch, line 114: 👍

apps/cli/src/store/local/actions.ts
```diff
@@ -333,31 +349,72 @@ export async function processProject(
...
+            ];
+          })
+          .reduce((a, b) => [a[0] + b[0], a[1] + b[1]], [0, 0]);
+      })
+      .reduce((a, b) => [a[0] + b[0], a[1] + b[1]], [0, 0]);
+    if (hasWarnings[0] > 0) {
+      const pluralE = hasWarnings[0] > 1 ? 's' : '';
+      const pluralW = hasWarnings[1] > 1 ? 's' : '';
+      throw new Error(
+        `Site has ${hasWarnings[0]} error${pluralE} and  ${hasWarnings[1]} warning${pluralW}, stopping build.`,
```
fwkoch:
```suggestion
        `Site has ${hasWarnings[0]} error${pluralE} and ${hasWarnings[1]} warning${pluralW}, stopping build.`,
```

apps/cli/src/store/local/actions.ts
```diff
@@ -333,31 +349,72 @@ export async function processProject(
...
-  await Promise.all(
+  const projects = await Promise.all(
     siteConfig.projects.map((siteProject) => processProject(session, siteProject, opts)),
   );
-  await writeSiteManifest(session);
-  // Copy all assets
-  copyLogo(session, siteConfig.logo);
-  siteConfig.actions.forEach((action) => copyActionResource(session, action));
+  if (opts?.strict) {
+    const hasWarnings = projects
```
fwkoch, line 386: Hah, the details of this block are super hard to parse - but I think that doesn't matter at all. It's just printing all warnings and errors, then throwing - I guess only if there are errors? Warnings are ok even for strict?

apps/cli/src/transforms/links.ts
```diff
@@ -96,13 +101,20 @@ function mutateStaticLink(session: ISession, link: GenericNode, linkFile: string
...
+
+export async function transformLinks(
+  session: ISession,
+  file: string,
+  mdast: Root,
+  opts?: { checkLinks?: boolean },
+): Promise<string[]> {
   const links = selectAll('link,linkBlock', mdast) as GenericNode[];
-  links.forEach((link) => {
+  links.map((link) => {
```
fwkoch: Just curious - what's the reasoning for this change? Naively, seems equivalent to me with in-place operations that don't return anything.

fwkoch (approved):
Overall looks great. The new error gathering framework is nice and quite reusable, and the link-check functions seem pretty simple and mostly in the right place.

Just a few secondary comments about smaller things.

apps/cli/src/cli/web.ts
rowanc1: I don't think so, we want the build/deploy to be permissive & fast by default.

apps/cli/src/transforms/images.ts
rowanc1: Fixed!

apps/cli/src/store/local/actions.ts
rowanc1: +1

apps/cli/src/transforms/links.ts
rowanc1: Changed it back.

rowanc1: Made this error in future:
image

apps/cli/src/store/build/actions.ts
rowanc1: moved.

PR #287, opened 2022-07-26
Update writeTOC to ignore hidden blocks

### LINKED ISSUE #203: only publish items that are marked as published on curvenote.com?

Yxwww:
- [x] test this change

### DIFF
--- a/apps/cli/src/export/jupyter-book/toc.ts
+++ b/apps/cli/src/export/jupyter-book/toc.ts
@@ -126,7 +126,7 @@ export async function writeTOC(session: ISession, nav: Version<Blocks.Navigation
       return;
     }
     const { parentId, block } = data;
-    if (!block) return;
+    if (!block || block.$data?.hidden) return;
     if (parentId) {
       const folder = nest[parentId];
       folder.push({ id, kind, block, children });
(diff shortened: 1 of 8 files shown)


apps/cli/src/export/jupyter-book/toc.ts
```diff
@@ -126,7 +126,7 @@ export async function writeTOC(session: ISession, nav: Version<Blocks.Navigation
       return;
     }
     const { parentId, block } = data;
-    if (!block) return;
+    if (!block || block.$data?.hidden) return;
```
rowanc1: Can use `block.data` which casts the data and fills in defaults.

fwkoch (commented):
We need to maybe be a little careful of edge cases, where hidden files may result in an invalid (e.g. empty, bad structured) toc. Maybe we throw a `validateTOC` at the end... but if it's invalid, what do we do? I guess just log an error and leave it at that.

apps/cli/src/sync/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface SyncCiHelperOptions {
```
fwkoch, line 1: 👍

fwkoch (commented):
Looks great to me (though a lot of this we paired on).

PR #288, opened 2022-07-26
🍽  Allow table of contents to be relative

rowanc1:
Previously the path would always be from the command location, so a _toc.yml in a nested directory would fail.

### DIFF
--- a/.changeset/polite-apricots-decide.md
+++ b/.changeset/polite-apricots-decide.md
@@ -1,0 +1,5 @@
+---
+'curvenote': patch
+---
+
+Relative links in `--write-toc` option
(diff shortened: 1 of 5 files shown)


fwkoch (commented):
Ok, is this addressing a bug where paths were wrong in the `_toc` if you were not in the same directory...? i.e. We are still writing the `_toc` in the root of the project (i.e. next to `curvenote.yml`) but now it works correctly regardless of the cwd?

rowanc1:
It writes in the project directory. e.g. you delete the one given and then say write-toc, and the paths were wrong, yeah.

PR #289, opened 2022-07-26
👩‍🔬  Review docs and update parameters

rowanc1:
Updates a number of things as I am going through the docs.

### DIFF
--- a/apps/cli/src/cli/utils.ts
+++ b/apps/cli/src/cli/utils.ts
@@ -7,6 +7,8 @@ import { chalkLogger, LogLevel } from '../logging';
 import { Session, getToken } from '../session';
 import { ISession } from '../session/types';
 import { selectors } from '../store';
+import CurvenoteVersion from '../version';
+import { packageJsonPath } from '../utils';
 
 const INSTALL_NODE_MESSAGE = `
 You can download Node here:
@@ -43,6 +45,22 @@ async function checkNodeVersion(session: ISession): Promise<boolean> {
     });
   });
   return checking;
+}
+
+function logVersions(session: ISession, debug = true) {
+  let siteVersion = '';
+  try {
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath(session)).toString()) as {
+      name: string;
+      version: string;
+    };
+    siteVersion = `\n - ${packageJson.name}: ${packageJson.version}`;
+  } catch (error) {
+    // pass
+  }
+  session.log[debug ? 'debug' : 'info'](
+    `\n\nVersions:\n - curvenote: ${CurvenoteVersion}${siteVersion}\n\n`,
+  );
 }
 
 type SessionOpts = {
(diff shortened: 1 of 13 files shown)


apps/cli/src/cli/utils.ts
```diff
@@ -116,12 +135,11 @@ export function clirun(
     try {
       await func(useSession, ...args);
     } catch (error) {
-      if (opts.debug === true) {
+      if (opts.debug) {
         useSession.log.debug(`\n\n${(error as Error)?.stack}\n\n`);
-      } else if (opts.debug) {
-        fs.writeFileSync(opts.debug, (error as Error)?.stack ?? '');
```
fwkoch, line 122: Is debug just always a bool now, no more debug file? I think that's totally fine, but we should change the type if so: https://github.com/curvenote/curvenote/pull/289#pullrequestreview-1051700437

apps/cli/src/utils/utils.ts
```diff
@@ -24,6 +24,10 @@ export function repoPath(session: ISession): string {
   return path.resolve(path.join('.', buildPath, 'curvenote'));
 }
 
+export function packageJsonPath(session: ISession): string {
```
fwkoch: A little ambiguous name... maybe 
```suggestion
export function webPackageJsonPath(session: ISession): string {
```
or something? That would make `logVersions` slightly more understandable, without reading this function, cli vs web build versions.

apps/cli/src/cli/utils.ts
```diff
@@ -7,6 +7,8 @@ import { chalkLogger, LogLevel } from '../logging';
 import { Session, getToken } from '../session';
 import { ISession } from '../session/types';
 import { selectors } from '../store';
+import CurvenoteVersion from '../version';
+import { packageJsonPath } from '../utils';
```
fwkoch:
```suggestion
import { webPackageJsonPath } from '../utils';
```

apps/cli/src/cli/utils.ts
```diff
@@ -45,6 +47,22 @@ async function checkNodeVersion(session: ISession): Promise<boolean> {
   return checking;
 }
 
+function logVersions(session: ISession, debug = true) {
+  let siteVersion = '';
+  try {
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath(session)).toString()) as {
```
fwkoch:
```suggestion
    const packageJson = JSON.parse(fs.readFileSync(webPackageJsonPath(session)).toString()) as {
```

fwkoch (commented):
Bunch of good straightforward changes.

apps/cli/src/cli/utils.ts
```diff
@@ -45,6 +47,22 @@ async function checkNodeVersion(session: ISession): Promise<boolean> {
...
+  } catch (error) {
+    // pass
+  }
+  session.log[debug ? 'debug' : 'info'](
+    `\n\nVersions:\n - curvenote: ${CurvenoteVersion}${siteVersion}\n\n`,
+  );
+}
+
 type SessionOpts = {
   debug?: boolean | string;
```
fwkoch:
```suggestion
  debug?: boolean;
```

PR #291, opened 2022-07-27
🌈 Look up all the references in the current tree

rowanc1:
Previously we would only load `references.bib` this now loads all references on the path.

### DIFF
--- a/apps/cli/src/toc/utils.ts
+++ b/apps/cli/src/toc/utils.ts
@@ -1,5 +1,5 @@
 import fs from 'fs';
-import { extname, join, parse } from 'path';
+import { extname, join, parse, basename } from 'path';
 import { title2name as createSlug } from '@curvenote/blocks';
 import { loadConfigOrThrow } from '../config';
 import { CURVENOTE_YML } from '../config/types';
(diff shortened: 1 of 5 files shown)


apps/cli/src/toc/utils.ts
```diff
@@ -77,14 +77,30 @@ export function fileInfo(file: string, pageSlugs: PageSlugs): { slug: string; ti
...
+  content
+    .map((dir) => join(path, dir))
+    .filter((file) => {
+      const isDir = isDirectory(file);
+      if (!isDir && extname(file) === '.bib') {
+        // Push the bibtex file to a list!
+        bibFiles.push(file);
+      }
+      // If it is in a list or is hidden
+      if (basename(file).startsWith('.') || BIBTEX_IGNORE_DIRECTORIES.includes(basename(file))) {
```
fwkoch: We should just reuse the `alwaysIgnore` function from `toc/fromPaths.ts`? Probably move that function to this file, and extend the ignore values to include those in `BIBTEX_IGNORE_DIRECTORIES` ... https://github.com/curvenote/curvenote/blob/main/apps/cli/src/toc/fromPath.ts#L18-L23

fwkoch (commented):
Love it!

apps/cli/src/toc/utils.ts
rowanc1: I was looking for that!

I think we can safely make them the same.

PR #298, opened 2022-08-05
⚙️ some cli testing on ci

stevejpurves:
possible new workflow to test cli installs over different OSs

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -1,10 +1,6 @@
 name: CLI
 
-on:
-  push:
-    branches: [ci-*]
-  pull_request:
-    branches: [main]
+on: workflow_dispatch
 
 jobs:
   cli:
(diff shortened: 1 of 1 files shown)


stevejpurves:
@rowanc1 blocks build fails on windows

rowanc1:
Hm, this is good, but shouldn't run on every commit because it is testing prod npm.

stevejpurves:
@rowanc1 ok we could remove the push to main trigger but on the pull_request all we can do there is not run it in draft mode? 


hm... wait...

stevejpurves:
ok, I have changed this to `workflow_dispatch` for now, which just basically means that we can only use it as a post-publish check on something already deployed.

To go beyond that do we need 2 workflows? a first build workflow in an OS we support builds for, which makes a build artifact available, then this workflow that then downloads that build artifact only and does `npm -g install /local-folder` on the OS matrix in order to test the cli install process?

Then we trigger that either manually, pre-publish or on commits to main

PR #304, opened 2022-08-08
👩‍🔬 Fetch citations from remote URL

rowanc1:
This allows for the bibliography list to be specified in the project curvenote.yml:

For example:

```
  bibliography:
    - references.bib
    - https://paperpile.com/eb/ySptJrCGMt/paperpile.bib
```

This will load the local references and join them with the remote references in order. And missing files will warn.

### DIFF
--- a/.changeset/green-pillows-march.md
+++ b/.changeset/green-pillows-march.md
@@ -1,0 +1,5 @@
+---
+'curvenote': patch
+---
+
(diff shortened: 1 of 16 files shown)


rowanc1:
Very strange, the tests were green when I hit merge. Fixed them up in dev, they were type errors.

PR #306, opened 2022-08-12
Enable builds via xelatex or pdflatex

stevejpurves:
This PR adds an additional option on `export pdf:build` enabling `pdflatex` to be used.


stevejpurves:
partially addresses #307

fwkoch:
This has been implemented as template option `build.engine` which jtex uses to construct the build command.

PR #309, opened 2022-08-16
Improve logging in a few areas

rowanc1:
image

### DIFF
--- a/apps/cli/src/cli/utils.ts
+++ b/apps/cli/src/cli/utils.ts
@@ -41,8 +41,7 @@ async function getNodeVersion(session: ISession): Promise<VersionResults | null>
   return result;
 }
 
-async function logVersions(session: ISession, debug = true) {
-  const result = await getNodeVersion(session);
+function logVersions(session: ISession, result: VersionResults | null, debug = true) {
   const versions: string[][] = [];
   Object.entries(result?.versions ?? {}).forEach(([name, p]) => {
     versions.push([
@@ -77,7 +76,8 @@ async function checkNodeVersion(session: ISession): Promise<boolean> {
   const result = await getNodeVersion(session);
   if (!result) return false;
   if (result.isSatisfied) return true;
-  await logVersions(session, false);
+  const versions = await getNodeVersion(session);
+  logVersions(session, versions, false);
   session.log.error('Please update your Node or NPM versions.\n');
   session.log.info(INSTALL_NODE_MESSAGE);
   return false;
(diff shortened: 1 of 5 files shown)


apps/cli/src/logging/custom.ts
```diff
@@ -34,7 +34,12 @@ export function getNpmLogger(session: ISession): LoggerDE {
     error(data: string) {
       const line = data.trim();
       if (!line) return;
-      if (line.includes('deprecated') || line === 'npm' || line.includes('WARN')) {
+      if (
+        line.includes('deprecated') ||
+        line.includes('package is no longer supported') ||
```
rowanc1, line 39: This silences the crypto deprecation warning

apps/cli/src/cli/utils.ts
```diff
@@ -159,7 +160,7 @@ export function clirun(
         useSession.log.debug(`\n\n${(error as Error)?.stack}\n\n`);
       }
       useSession.log.error((error as Error).message);
-      await logVersions(useSession, false);
+      logVersions(useSession, versions, false);
```
rowanc1, line 163: logVersions is now not async, which was giving challenges in showing the error message in a different place than the process.exit when multiple things were running.

apps/cli/src/licenses/validators.ts
```diff
@@ -84,12 +85,16 @@ export function validateLicense(input: any, opts: Options): License | undefined
...
   }
-  let value = validateString(input, opts);
-  if (value === undefined) return undefined;
-  value = value.toUpperCase();
+  const valueUnvalidated = validateString(input, opts);
+  if (valueUnvalidated === undefined) return undefined;
+  const value = valueUnvalidated.toUpperCase();
   if (!LICENSE_KEYS[value]) {
+    const possibleAlts = Object.keys(LICENSE_KEYS).filter((k) => k.includes(value));
+    const alts = possibleAlts.length > 0 ? ` Maybe try:\n- "${possibleAlts.join('"\n- "')}"\n` : '';
```
rowanc1, line 93: This was helpful for things like GPL, or CC-BY

apps/cli/src/toc/fromToc.ts
```diff
@@ -51,7 +51,13 @@ export function projectFromToc(
   const pageSlugs: PageSlugs = {};
   const indexFile = resolveExtension(join(path, toc.root));
   if (!indexFile) {
-    throw Error(`Root from ${tocFile(path)} not found: ${indexFile}`);
+    throw Error(
```
rowanc1, line 54: @stevejpurves this was the error hit by one of our users recently!

fwkoch (approved):

apps/cli/src/logging/custom.ts
fwkoch: Good to silence this for users. Should we resolve it at some point though? Not sure if there's another reason to stick with deprecated package.

rowanc1: It is baked into some Jupyter dependencies, so yes, but might take a while for updates to happen.

PR #316, opened 2022-08-22
🐛  Frontmatter is deleted if it is the only node

rowanc1:
If the content is only a frontmatter block, then the `remove` tries to delete the root node, which is not allowed. This follows through with removing that content.

Content:
image

This is the bug:
image

After the content is properly deleted and a blank article is shown.

### DIFF
--- a/apps/cli/src/frontmatter/index.ts
+++ b/apps/cli/src/frontmatter/index.ts
@@ -1,5 +1,5 @@
 import yaml from 'js-yaml';
-import type { Root, Heading } from 'mdast';
+import type { Root, Code, Heading } from 'mdast';
 import { remove, select } from 'mystjs';
 import type { Licenses } from '../licenses/types';
 import { licensesToString } from '../licenses/validators';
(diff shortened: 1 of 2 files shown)


apps/cli/src/frontmatter/index.ts
```diff
@@ -42,7 +42,11 @@ export function frontmatterFromMdastTree(
   }
   if (removeNode) {
     // Handles deleting the block if it is the only element in the block
-    remove(tree, '__delete__');
+    const possibleNull = remove(tree, '__delete__');
+    if (possibleNull === null) {
+      // null is returned if tree itself didn’t pass the test or is cascaded away
+      remove(tree, { cascade: false }, '__delete__');
```
rowanc1, line 48: See, a bit surprising, but makes sense.

https://github.com/syntax-tree/unist-util-remove#removetree-options-test

fwkoch (approved):
This makes sense, previously was cascading up, nuking the root, and leaving nothing. Now it will leave empty children on the root.

PR #317, opened 2022-08-23
🚀 Initial myst transforms

rowanc1:
This starts to pull the transforms used in the CLI into a place that can be shared with other applications, notably the demo app.

- [x] section number tests (enumerate)
- [x] what to do with IDs that are not HTML worthy?
- [x] use math transform
- [x] add file warnings infrastructure, move this over in the CLI
- [x] move more transforms over into shared package (code, imageAlt, blockNesting)
- [x] Basic docs on each transform

Features:
https://twitter.com/rowancockett/status/1562133043296382976


rowanc1:
This improves the section reference lookup:

image

rowanc1:
- [ ] section number tests (enumerate)
- [ ] what to do with IDs that are not HTML worthy?
- [ ] use math transform
- [ ] add file warnings infrastructure, move this over in the CLI
- [ ] move more transforms over into shared package (code, imageAlt, blockNesting)

rowanc1:
Merging this in will break any CLI calls that pull from curvenote for a new website. That is because we had to change the build process, and the CLI just pulls from main, not a versioned release.

If we follow this up with a release, I think that is fine for now.

PR #321, opened 2022-09-09
Feat/tex word references

rowanc1:
@fwkoch this is me rebasing so we can merge, can you check the changes between the branches, and if this is the more up to date one we can use it to merge into main.


rowanc1:
VS code changed their UI and now I basically can't use git ... :)

rowanc1:
Taken over in #322

PR #323, opened 2022-09-09
🎇 Improve mdast image paths

fwkoch:
Pass `imageWriteFolder` (and optionally `imageAltOutputFolder`) to `transformMdast`

`imageWriteFolder` is the local folder where image is written; if alt output folder is provided, that is written to the transformed mdast. This is used when building a site to save images to a build folder but writing a static web path.


rowanc1:
Found a bug -- need to ensure the write folder exists!

PR #325, opened 2022-09-12
Improve `jtex` docs and create a CLI

rowanc1:
This is a first version of a CLI that was helping me in validating some of the templates, I already found some errors/types with this, and it is a very easy way to validate a template without having to render it.

image

There is also a first pass of listing from the API.

image

In the docs, I have gone with the `parts` terminology over tagged. @fwkoch curious if you are game to just make this change?

I also think that flattening the YAML slightly, by removing `metadata` and `config` and having their children be top level?


packages/jtex/src/cli/check.ts
```diff
@@ -0,0 +1,240 @@
...
+    });
+  }
+
+  // Validate non-frontmatter options
+  const doc = validated.options?.filter((p) => p.type === 'frontmatter').map((p) => p.id) ?? [];
+  Object.entries(variables.doc).forEach(([optKey, lineNumbers]) => {
+    if (!doc.includes(optKey) && DOC_FRONTMATTER_KEYS.includes(optKey)) {
+      messages.errors.push({
+        property: 'doc',
+        message: `The template.yml does not frontmatter option "${optKey}" but it is referenced in template.tex on ${lineNumbersToString(
```
fwkoch:
```suggestion
        message: `The template.yml does not include document property "${optKey}" but it is referenced in template.tex on ${lineNumbersToString(
```

packages/jtex/src/cli/check.ts
```diff
@@ -0,0 +1,240 @@
...
+
+  // Validate parts
+  const parts = validated.tagged?.map((p) => p.id) ?? [];
+  const usedParts: string[] = [];
+  Object.entries(variables.parts).forEach(([partKey, lineNumbers]) => {
+    if (!parts.includes(partKey)) {
+      const lines = `line${lineNumbers.length > 1 ? 's' : ''} ${lineNumbers.join(', ')}`;
+      messages.errors.push({
+        property: 'parts',
+        message: `The template.yml does not include "${partKey}" but it is referenced in template.tex on ${lines}`,
```
fwkoch:
```suggestion
        message: `The template.yml does not include part "${partKey}" but it is referenced in template.tex on ${lines}`,
```

packages/jtex/src/cli/check.ts
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  // Validate non-frontmatter options
+  const options = validated.options?.filter((p) => p.type !== 'frontmatter').map((p) => p.id) ?? [];
+  const usedOpts: string[] = [];
+  Object.entries(variables.options).forEach(([optKey, lineNumbers]) => {
+    if (!options.includes(optKey)) {
+      messages.errors.push({
+        property: 'options',
+        message: `The template.yml does not include "${optKey}" but it is referenced in template.tex on ${lineNumbersToString(
```
fwkoch:
```suggestion
        message: `The template.yml does not include option "${optKey}" but it is referenced in template.tex on ${lineNumbersToString(
```

packages/jtex/src/cli/check.ts
```diff
@@ -0,0 +1,240 @@
...
+  });
+  messages.errors?.forEach((e) => {
+    session.log.error(`  [${e.property}] ${e.message}`);
+  });
+  messages.warnings = [];
+  messages.errors = [];
+  return true;
+}
+
+export function checkTemplate(session: ISession, path: string) {
```
fwkoch, line 104: Wondering if this should also check that `[-CONTENT-]` is present?

rowanc1: Yep, and imports, good call.

rowanc1: image

PR #327, opened 2022-09-13
🍽 don't nest nav blocks with missing parents

stevejpurves:
We have observed that a nav item at root level can have a `parentId` of a non-existent nav block. We're not clear on how to get into that state, but the change in this PR protects against it on building a TOC during the site build process


stevejpurves:
ok, you can get into this state by deleting an article in the app, that has other articles nested under it.
Deleting a group clears up the parentId field, but deleting an article does not.

stevejpurves:
Did some more testing on this, this morning with some different projects. Seems good.

PR #330, opened 2022-09-14
Updates to template yml

fwkoch:
- [x] export -> exports
- [x] Add template/options to export type
- [x] Respect template/output (folder or file) in frontmatter when exporting
- [x] Reserve these keywords ^ so options cannot clobber
- [x] str/bool -> string/boolean for option types
- [x] Move frontmatter options out to doc
- [x] Flatten metadata, schema on template yml (and add more validation to metadata fields)
- [x] Add packages to template yml


packages/frontmatter/src/frontmatter/validators.ts
```diff
@@ -101,6 +101,7 @@ const NUMBERING_KEYS = [
 const KERNELSPEC_KEYS = ['name', 'language', 'display_name', 'argv', 'env'];
 const TEXT_REPRESENTATION_KEYS = ['extension', 'format_name', 'format_version', 'jupytext_version'];
 const JUPYTEXT_KEYS = ['formats', 'text_representation'];
+export const RESERVED_EXPORT_KEYS = ['format', 'template', 'output'];
```
rowanc1: maybe `name` and `id` as well??

Issue #307, opened 2022-08-12
Enable automatic export to pdf using templates that require pdflatex


stevejpurves:
Some LaTeX templates contain `pdflatex` specific macros and must be built using that build command. For example, EGU Copernicus https://github.com/curvenote/templates1/issues/4 requires this.

PR #306 adds the correct build command to run pdflatex

In order to make this possible on a per template basis and enable builds from curvenote.com, we need to:

 - [ ] Extend the `template.yml` schema, currently defined in the `jtex` package to include a field such as `build` or `engine` which can be `xelatex` or `pdflatex`
 - [ ] in the Curvenote CLI we already fetch the template spec (`template.yml`) in order to determine the tagged blocks required by the template. We should refactor and move that call so that the `engine` option can be used appropriately in the build pdf function call

rowanc1:
This is now complete! EGU builds well!

PR #332, opened 2022-09-15
📍 Local PDF export using local tex export

fwkoch:
- [x] Command line pdf export works with local file and uses frontmatter options
- [x] Logging messages on pdf build
- [x] Use format `pdf+tex` to keep build artifacts
- [x] Allow alternative build engine `pdflatex`: https://github.com/curvenote/curvenote/issues/307

More things:
- [x] Don't fall back to project `exports` on page frontmatter (we maybe want to do something more complicated like pages can use the `template` specified in project frontmatter... but simply populating pages with project `exports` feels wrong)
- [x] Fill frontmatter from project (e.g. authors)
- [x] Default `_build` directory should be in project root

### DIFF
--- a/packages/jtex/src/export.ts
+++ b/packages/jtex/src/export.ts
@@ -1,0 +1,4 @@
+export function pdfExportCommand(texFile: string, logFile: string, engine?: string) {
+  if (!engine) engine = 'xelatex';
+  return `latexmk -f -${engine} -synctex=1 -interaction=batchmode -file-line-error -latexoption="-shell-escape" ${texFile} &> ${logFile}`;
+}
--- a/apps/cli/src/cli/export.ts
+++ b/apps/cli/src/cli/export.ts
@@ -9,6 +9,7 @@ import {
   oxaLinkToPdf,
   pathToWord,
 } from '../export';
+import { makeCleanOption } from './options';
 import { clirun } from './utils';
 
 function makeImageOption() {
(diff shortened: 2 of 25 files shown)


rowanc1:
Performance on the docs:

```
Pre:
curvenote build -c  51.58s user 3.25s system 358% cpu 15.279 total
curvenote build  1.88s user 0.22s system 132% cpu 1.588 total

Image MD5:
curvenote build -c  65.26s user 3.77s system 250% cpu 27.586 total
curvenote build  17.38s user 0.78s system 108% cpu 16.722 total

Image date/size
curvenote build -c  49.73s user 2.96s system 369% cpu 14.275 total
curvenote build  1.67s user 0.20s system 133% cpu 1.401 total
```

So this is (maybe) slightly faster, but at least not 8x slower. :)

rowanc1:
Tested out functionality and all looks really good, no unexpected issues found, only major thing is copying in the bibtex file(s).

Lets get this in first thing tomorrow!

packages/jtex/src/export.ts
```diff
@@ -0,0 +1,4 @@
+export function pdfExportCommand(texFile: string, logFile: string, engine?: string) {
+  if (!engine) engine = 'xelatex';
```
rowanc1, line 2: Just making sure you know about this syntax @fwkoch:
```suggestion
export function pdfExportCommand(texFile: string, logFile: string, engine = 'xelatex') {
```

PR #345, opened 2022-09-19
Use only SiteConfig type/validation defined in blocks

fwkoch:


apps/cli/src/toc/types.ts
```diff
@@ -59,8 +59,8 @@ export type SiteManifest = {
   title: string;
   twitter?: string;
   logo?: string;
-  logoText?: string;
-  nav: SiteNavItem[];
+  logo_text?: string;
```
rowanc1: Renaming to logo_text?

apps/cli/src/sync/clone.ts
```diff
@@ -97,10 +98,13 @@ export async function clone(session: ISession, remote?: string, path?: string, o
     );
     writeConfigs(session, '.', { siteConfig: newSiteConfig });
   } else if (siteConfig) {
-    const newSiteConfig: SiteConfig = {
+    const newSiteConfig: PartialSiteConfig = {
```
rowanc1: I think that we likely want to clean up the names of these in the next pass, but that can be a refactor.

PartialSiteConfig, which is missing the ID, date_create etc, likely isn't what we want from the myst side.

package-lock.json
```diff
@@ -31574,7 +31576,7 @@
         "@types/react": "^17.0.37",
         "@types/react-dom": "^17.0.11",
         "ansi-to-react": "^6.1.6",
-        "buffer": "*",
+        "buffer": "^6.0.3",
```
rowanc1, line 31579: Thanks.

packages/blocks/src/sites.ts
```diff
@@ -85,39 +90,35 @@ export interface SiteAnalytics {
...
-    'description',
-    'venue',
     'projects',
     'nav',
     'actions',
+    'domains',
     'twitter',
     'logo',
     'logo_text',
+    'logoText',
```
rowanc1:
```suggestion
    'logoText', // Deprecated, but included in validation
```


I am actually not sure if this should be in here? Shouldn't we warn on it?

rowanc1 (commented):
Looks good. One question on the validation.

packages/blocks/src/sites.ts
fwkoch: We filter against these keys before we do any other validation: https://github.com/curvenote/curvenote/pull/345/files/62970e02c90ffce8046149cea36af5081ee7e112#diff-01b0951fb3833dadc72bbf40ff110278796d45b9282ce0487f68b68775785ff2R372

I suppose it would make sense to do this immediately on `readConfig`, the same place where we un-nest frontmatter and warn...

apps/cli/src/sync/clone.ts
fwkoch: Hmm, I think we want `blocks.PartialSiteConfig` to become `myst.SiteConfig`, and then we have an extended `blocks.SiteConfigDTO` or something with `id`, etc...

Maybe worth a small rename now, so we aren't using "PartialSiteConfig" which sounds sketchy.

package-lock.json
fwkoch: 🤔 Hah, this just happened on it's own I guess?

PR #351, opened 2022-09-21
Feat/landing

rowanc1:


packages/frontmatter/src/frontmatter/types.ts
```diff
@@ -93,11 +93,20 @@ export type Export = {
...
+  hide_outline?: boolean;
+  hide_footer_links?: boolean;
+};
+
 export type SiteFrontmatter = {
   title?: string;
   description?: string;
   venue?: Venue;
   keywords?: string[];
+  design?: SiteDesign;
```
fwkoch, line 109: Probably should remove `design` from `SiteManifest` if we are adding it here, just to not have the duplication.

packages/frontmatter/src/frontmatter/frontmatter.spec.ts
```diff
@@ -192,6 +193,18 @@ describe('validateBiblio', () => {
   });
 });
 
+describe('validateSiteDesign', () => {
+  it('empty object returns self', async () => {
+    expect(validateSiteDesign({}, opts)).toEqual({});
+  });
+  it('valid site design returns self', async () => {
+    const siteDesign = {
+      hide_authors: true,
```
fwkoch, line 202: Should chuck all fields on here.

packages/frontmatter/src/frontmatter/validators.ts
```diff
@@ -53,6 +54,7 @@ export const PROJECT_FRONTMATTER_KEYS = [
   'bibliography',
   'math',
   'exports',
+  'design',
```
fwkoch, line 57: This should be on `SITE_FRONTMATTER_KEYS`

packages/frontmatter/src/frontmatter/types.ts
fwkoch: Oh - I think the only reason this works is because `design` was not removed from `SiteManifest`... otherwise `SITE_FRONTMATTER_KEYS` without design would filter it out...

packages/frontmatter/src/frontmatter/validators.ts
```diff
@@ -53,6 +54,7 @@ export const PROJECT_FRONTMATTER_KEYS = [
...
 ].concat(SITE_FRONTMATTER_KEYS);
 export const PAGE_FRONTMATTER_KEYS = [
```
fwkoch, line 59: If we want to be able to define these on the page and override appropriately, we need to add `design` to `USE_SITE_FALLBACK`

fwkoch (commented):
A few funny things with this - looks like page-level frontmatter `design` isn't supported yet? A bit of rearranging may be necessary if we want all that to work smoothly.

However, if it works, that's probably ok for now.

PR #356, opened 2022-09-26
🌅  Write intersphinx references

rowanc1:
Intersphinx CLI:

image


packages/myst-cli-utils/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "description": "Utils for logging, error messages and session management",
+  "author": "Rowan Cockett <rowan@curvenote.com>",
+  "homepage": "https://github.com/curvenote/curvenote#readme",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
+  "types": "./dist/types/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "exports": {
```
fwkoch, line 14: We had to take this block out of `myst-common` to get builds working

apps/cli/src/store/local/actions.ts
```diff
@@ -448,6 +451,48 @@ export async function writeSiteManifest(session: ISession) {
...
+  session: ISession,
+  inv: Inventory,
+  opts: { projectPath: string },
+) {
+  const { pages } = loadProject(session, opts.projectPath);
+  const pageReferenceStates = selectPageReferenceStates(session, pages);
+  pageReferenceStates.forEach((page) => {
+    const { title } = selectors.selectFileInfo(session.store.getState(), page.file);
+    inv.setEntry({
+      type: 'std:doc',
```
fwkoch: Looks like these types are from `intersphinx` - `enum` maybe?

fwkoch (approved):
Not a super detailed review, but everything in here makes sense, the content of `objects.inv` seems straightforward.

apps/cli/src/store/local/actions.ts
rowanc1: Done.

packages/myst-cli-utils/package.json
rowanc1: Will leave it in and note it for later?

PR #357, opened 2022-09-26
📖 Copy bib files for tex export

fwkoch:


apps/cli/src/store/local/actions.ts
```diff
@@ -611,3 +616,8 @@ export async function processSite(session: ISession, opts?: ProcessOptions): Pro
   }
   return true;
 }
+
+export function bibtexSomething(session: ISession, path: string) {
+  const cache = castSession(session);
+  console.log(cache.$externalReferences);
+}
```
rowanc1: Not sure if this is left over?

fwkoch: Haha ok. Hmmf

packages/jtex/src/jtex.ts
```diff
@@ -145,6 +150,7 @@ class JTex {
       errorLogFn: errorLogger(this.session),
       warningLogFn: warningLogger(this.session),
     };
+    if (bibliography) frontmatter.bibliography = bibliography;
```
rowanc1: should we also clone this object? Probably not a big deal for our stuff, but probably best if we are making changes to a different doc.

packages/jtex/src/jtex.ts
```diff
@@ -145,6 +150,7 @@ class JTex {
...
     const validatedDoc = validateTemplateDoc(frontmatter, templateYml?.doc || [], options, opts);
```
rowanc1: Can you just check if we should be destructuring frontmatter before changing the bib?
ie.
```
const validatedDoc = validateTemplateDoc({...frontmatter, bibliography: bibliography ?? frontmatter.bibliography }, templateYml?.doc || [], options, opts);
 ```

packages/jtex/src/jtex.ts
fwkoch: I think the validation functions clone objects, but I did destructure before changing bib.

rowanc1: Nice!

PR #358, opened 2022-09-28
JTeX updates

rowanc1:
MyST templates now live here:
https://github.com/myst-templates


packages/myst-cli-utils/src/exec.ts
```diff
@@ -0,0 +1,39 @@
+import util from 'util';
+import child_process from 'child_process';
+import type { Logger } from './types';
+
+function execWrapper(
```
rowanc1, line 5: These could now be removed from the CLI, currently duplicated this file.

packages/myst-cli-utils/src/logger.ts
```diff
@@ -47,3 +47,50 @@ export function chalkLogger(level: LogLevel): Logger {
...
+        session.log.debug(line);
+        return;
+      }
+      session.log.error(data);
+    },
+  };
+  return logger;
+}
+
+export function createNpmLogger(session: ISession): LoggerDE {
```
rowanc1, line 92: These can be pulled out of the CLI as well.

packages/myst-to-tex/src/index.ts
```diff
@@ -296,14 +309,18 @@ class TexSerializer implements ITexSerializer {
...
 const plugin: Plugin<[Options?], Root, VFile> = function (opts) {
   this.Compiler = (node, file) => {
     const state = new TexSerializer(file, opts ?? { handlers });
     state.renderChildren(node);
     const tex = (file.result as string).trim();
     const result: LatexResult = {
-      imports: [],
-      commands: [...createMathCommands(state.mathPlugins)],
+      imports: createImportCommands(state.data.imports),
+      commands: createMathCommands(state.data.mathPlugins),
```
rowanc1, line 323: There are improvements here to make to do this data --> string conversion later after all parts are combined.

rowanc1 (commented):
Some todos!

PR #359, opened 2022-09-28
📬 Pull md then use local export for docx/pdf/tex remote export

fwkoch:


apps/cli/package.json
```diff
@@ -57,6 +57,7 @@
     "@google-cloud/storage": "^5.18.2",
     "@jupyterlab/nbformat": "^3.3.2",
     "@reduxjs/toolkit": "^1.7.2",
+    "adm-zip": "^0.5.9",
```
fwkoch, line 60: Don't love adding this dependency, but what we have in jtex is `unzipper` - not great for zipping, hah. Could potentially replace that with this...?

apps/cli/src/export/tex/single.ts
```diff
@@ -308,39 +267,59 @@ export async function runTexExport(
...
+  zip.writeZip(zipOutput);
+}
+
+export async function localArticleToTex(
+  session: ISession,
+  file: string,
+  opts: TexExportOptions,
+  templateOptions?: Record<string, any>,
+) {
+  console.log(templateOptions);
```
rowanc1: Remove me!
```suggestion
```

apps/cli/src/cli/export.ts
```diff
@@ -54,7 +50,7 @@ function makeWordExportCLI(program: Command) {
     .argument('<article>', 'A link to the Curvenote article (e.g. OXA Link or API link)')
     .argument('[output]', 'The document filename to export to', '')
     .addOption(makeCleanOption())
-    .action(clirun(pathToWord, { program }));
+    .action(clirun(pathToWord, { program }, 3));
```
rowanc1, line 53: What does the 3 do here?

rowanc1: Ah, I see, slicing args. Thought it was on the other side of the bracket!

fwkoch: Yeah, commander passes some giant object as the last argument. Cutting that off so it isn't passed to our functions...

rowanc1:
Looks like content is always nested under a `_tex` folder instead of being in the exports folder. Not sure if that was introduced in this PR, but I know we fixed that recently for the tex exports to be put in a folder.

image

PR #360, opened 2022-09-30
⚒️ Import latex packages

rowanc1:


packages/jtex/src/imports.ts
```diff
@@ -11,10 +11,37 @@ function label(title: string, commands: string[]) {
...
+export function renderImports(
+  expandedImports?: string | ExpandedImports,
+  existingPackages?: string[],
+): string {
   if (!expandedImports || typeof expandedImports === 'string') return expandedImports || '';
-  const imports = label('imports', [...new Set(expandedImports.imports)]);
-  const commands = label('math commands', [...new Set(expandedImports.commands)]);
+  const packages = new Set(expandedImports.imports);
+  const math = Object.fromEntries(
+    Object.entries(expandedImports.commands).sort(([[a], [b]]) => {
```
fwkoch: Should this sorting be in `createMathCommands` (like sorting is in `createImportCommands`) - getting a little bogged down, but I think this is just sorting to make the output nicer?

fwkoch (approved):

packages/jtex/src/imports.ts
rowanc1: Removed!

PR #366, opened 2022-10-07
📑 Move template yml types from jtex to myst-templates

fwkoch:
This allows us to import these types without all the additional jtex baggage.


rowanc1:
Can you add a changeset so we remember to bump both packages?

fwkoch:
This has been replaced with:
https://github.com/executablebooks/mystjs/pull/67
and
https://github.com/curvenote/curvenote/pull/387

PR #375, opened 2022-10-19
💥 Split myst-cli out from curvenote cli

fwkoch:
See: https://github.com/executablebooks/meta/issues/838


.changeset/smooth-monkeys-fetch.md
```diff
@@ -0,0 +1,6 @@
+---
+'curvenote': patch
```
fwkoch, line 2: Ugh, meant for this to be minor, but I guess I hit `enter` instead of `space` in the cli.

But I don't think it maters since there is another minor bump in this changeset.

rowanc1: Yep, we can fix it up and ensure that these are correct when we bump.

PR #377, opened 2022-10-19
🆕 Use the theme server

rowanc1:
This pulls out the theme from inside of the server, and sets up a standalone server with reload socket.

See https://github.com/curvenote/book-theme/issues/1


rowanc1:
Choosing to rebase this onto @fwkoch's myst-cli branch!

PR #378, opened 2022-10-20
🆕 Use the theme server

rowanc1:
Replaces #377

### DIFF
--- a/apps/cli/src/cli/utils.ts
+++ b/apps/cli/src/cli/utils.ts
@@ -2,7 +2,7 @@ import fs from 'fs';
 import chalk from 'chalk';
 import type check from 'check-node-version';
 import type { Command } from 'commander';
-import { getNodeVersion, selectors } from 'myst-cli';
+import { getNodeVersion, selectors, version as mystCliVersion } from 'myst-cli';
 import { chalkLogger, LogLevel } from 'myst-cli-utils';
 import { Session, getToken } from '../session';
 import type { ISession } from '../session/types';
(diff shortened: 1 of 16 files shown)


apps/cli/src/cli/utils.ts
```diff
@@ -35,13 +35,17 @@ function logVersions(session: ISession, result: VersionResults | null, debug = t
       p.isSatisfied ? '✅' : '⚠️',
     ]);
   });
+  versions.push(['myst-cli', mystCliVersion]);
```
fwkoch, line 38: 👍

apps/cli/src/web/deploy.ts
```diff
@@ -15,71 +15,31 @@ import type {
...
-function listPublic(session: ISession): FromTo[] {
-  const staticFolder = path.join(session.publicPath(), '_static');
-  if (!fs.existsSync(staticFolder)) return [];
-  const assets = fs.readdirSync(staticFolder);
-  const fromTo = assets.map((assetName) => {
-    return {
-      from: path.join(staticFolder, assetName),
-      to: `public/_static/${assetName}`,
-    };
+function listFolderContents(session: ISession, from: string, to = ''): FromTo[] {
```
fwkoch, line 24: Seems simpler, assuming we aren't losing anything with, like the default logo or anything.

apps/cli/src/web/index.ts
```diff
@@ -52,10 +61,10 @@ export async function install(session: ISession): Promise<void> {
   }
   await makeExecutable('npm install', createNpmLogger(session), { cwd: session.repoPath() })();
   session.log.info(toc('📦 Installed web libraries in %s'));
-  await makeExecutable('npm run build:web', createNpmLogger(session), {
-    cwd: session.repoPath(),
-  })();
-  session.log.info(toc('🛠  Built dependencies in %s'));
+  // await makeExecutable('npm run build:web', createNpmLogger(session), {
```
fwkoch, line 64: Just kill this commented stuff now?

fwkoch (approved):

apps/cli/src/web/deploy.ts
rowanc1: Nope, it was super complex before because it was all scattered. Now what is in site is also on the CDN.

This will also make it easier to copy in other content into this folder as well.

apps/cli/src/web/index.ts
rowanc1: I left it in because we will need to build in future, but that build step is dependent on the template.

PR #389, opened 2022-11-11
Add `parts` to block type

fwkoch:
This adds `parts` alongside `tags` on the `block` type. It also adds `parts` as myst metadata when pulling a block from the API.


rowanc1:
I am taking this over in #390

PR #390, opened 2022-11-11
💇‍♀️ Include parts in myst block metadata

rowanc1:


packages/blocks/src/blocks/types/misc.ts
```diff
@@ -20,17 +20,21 @@ export enum XClientName {
 
 export const TAG_ABSTRACT = 'abstract';
 
-export enum WellKnownBlockTags {
+export enum WellKnownBlockParts {
```
fwkoch, line 23: 👍

PR #394, opened 2022-11-14
🐛 Remove setting of nav on init

rowanc1:
This no longer creates a navigation list on `curvenote init`. This often led to a duplicate title and first nav item.

### DIFF
--- a/packages/curvenote/src/sync/init.ts
+++ b/packages/curvenote/src/sync/init.ts
@@ -4,7 +4,7 @@ import inquirer from 'inquirer';
 import { basename, join, resolve } from 'path';
 import { config, findProjectsOnPath, loadProjectFromDisk, selectors, writeConfigs } from 'myst-cli';
 import { LogLevel, writeFileToFolder } from 'myst-cli-utils';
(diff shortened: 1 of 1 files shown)


packages/curvenote/src/sync/init.ts
```diff
@@ -150,27 +150,11 @@ export async function init(session: ISession, opts: Options) {
...
-          title: projConf?.title || proj.slug,
-          url: `/${proj.slug}`,
-        };
-      } else if (proj.remote) {
-        return { title: proj.slug, url: proj.remote };
-      } else {
-        return undefined;
-      }
-    })
-    .filter((proj): proj is SiteNavPage => Boolean(proj));
```
stevejpurves, line 173: @rowanc1 so now nav creation will happen when? on load in the theme? or essentially by default no nav is created no matter how many projects are in the site?

rowanc1: This is just the init function for creating the default nav. The main way it is used isn't that helpful ... It just duplicates the title. Removing that means the defaults are visually better, but you have to go to the docs to learn how to put a nav item in.

PR #427, opened 2023-06-27
🐛 Fix export wrapper bugs and pass through temp folders / log files

fwkoch:
This resolves failures when exporting from a curvenote link - the project path was not maintained, so source md files were not found correctly.

This also consumes https://github.com/executablebooks/mystjs/pull/442 to pass through temp folders / log files created during exports.

### DIFF
--- a/.changeset/neat-dolls-destroy.md
+++ b/.changeset/neat-dolls-destroy.md
@@ -1,0 +1,5 @@
+---
+'curvenote': patch
+---
+
+Fix project path when exporting from a curvenote url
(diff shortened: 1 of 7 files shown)


stevejpurves:
just tested the single export from link command i.e.
```
curvenote export pdf https://curvenote.com/@stevejpurves/la-palma-earthquakes/la-palma-eruption-2022-paper
```

which is working well!

rowanc1:
How about we release myst and then get this in?

PR #433, opened 2023-08-02
🛤 Track myst-cli v1.1.8

rowanc1:


packages/curvenote/src/frontmatter/api.ts
```diff
@@ -39,21 +39,25 @@ export function projectFrontmatterFromDTO(
 ): ProjectFrontmatter {
   const apiFrontmatter = filterKeys(project, PROJECT_FRONTMATTER_KEYS) as ProjectFrontmatter;
   if (apiFrontmatter.authors) {
-    apiFrontmatter.authors = apiFrontmatter.authors.map((author: Author) =>
-      resolveAffiliations(session, author),
-    );
+    apiFrontmatter.authors = apiFrontmatter.authors.map((author: Author) => {
+      const resolvedAuthor = resolveAffiliations(session, author);
+      delete resolvedAuthor.id;
```
fwkoch, line 44: The changes in this file back out new additions from the updated affiliations in `myst-frontmatter`. We need to take a step where we reconcile affiliations in the API / CLI store with the new `Affiliations` object introduced here: https://github.com/executablebooks/mystmd/pull/535

Until then, though, this keeps things simple going between API affiliations and MyST frontmatter. Affiliations come through as simple strings still, and those are still handled nicely by the updated `myst-frontmatter` validation.

PR #434, opened 2023-08-22
✅ Add framework and examples for manuscript checks

fwkoch:
image

This PR provides an interface for defining arbitrary checks against a MyST manuscript. It also provides a function to run these checks.

For now there are a couple example checks implemented here and called from the CLI; however, these checks are not yet extensible from the CLI.


packages/curvenote/src/check/examples/abstractLength.ts
```diff
@@ -0,0 +1,42 @@
...
+  id: 'abstract-length',
+  title: 'Abstract Length',
+  description: 'Ensure abstract in MyST project is the correct length',
+  category: 'abstract',
+  options: [
+    {
+      id: 'max',
+      title: 'Maximum word count for abstract',
+      required: true,
+      type: TemplateOptionType.string,
```
fwkoch, line 19: Todo: add `TemplateOptionType.number` as a valid `type` upstream.

packages/curvenote/src/check/runner.ts
```diff
@@ -0,0 +1,123 @@
...
+import { validateCheck } from './validators.js';
+
+export async function runChecks(
+  session: ISession,
+  file: string,
+  checks: Check[],
+  checkDefinitions: CheckInterface[],
+): Promise<CheckReport> {
+  const opts = { property: 'checks', messages: {} };
+  if (!['.ipynb', '.md', '.tex'].includes(path.extname(file))) {
```
fwkoch, line 30: Here we probably also want to support jats/meca/docx(?) - but I think those should likely be all loaded to mdast here, then the `check` functions don't really care about the source format, only the mdast.

That might be too idealistic though... we will have to see as we implement more real checks against real files.

stevejpurves:
Looks great! these are just unordered thoughts as I'm looking through things.

So the checks are only carried out on the file that is specified in the command? although the whole project is built and available. 

I guess an interesting check with much larger scope than the current example would be for something like cross references or embeds/figures where we need to be able to check a reference for a valid target in some other mdast file. 

Checking for citations also, requires interaction outside of the single file / mdast tree

packages/curvenote/src/check/examples/abstractLength.ts
```diff
@@ -0,0 +1,42 @@
...
+  validate: async (session: ISession, file: string, options: Check) => {
+    const { mdast } = selectFile(session, path.resolve(file)) ?? {};
+    if (!mdast) {
+      return { status: CheckStatus.error, message: `Error loading content from ${file}` };
+    }
+    const abstract = extractPart(copyNode(mdast), 'abstract');
+    if (!abstract) {
+      return { status: CheckStatus.error, message: `No abstract found in ${file}` };
+    }
+    const splitAbstract = toText(abstract).split(/\s+/);
```
rowanc1: We have used this package before:
https://www.npmjs.com/package/@wordpress/wordcount

rowanc1 (commented):
Exciting to see these come into existence!

packages/curvenote/src/check/examples/abstractLength.ts
rowanc1: Probably the same as default though.. !
https://github.com/WordPress/gutenberg/blob/f67c38677442518e65e338849eaa3fc435773a5c/packages/wordcount/src/defaultSettings.js#L96

stevejpurves (approved):

PR #440, opened 2023-09-07
extend the cli to enable private deployment and submission to venues

stevejpurves:
This is a first cut at enabling private uploads and venue submissions. The cli is doing some things to start with, that probably should be moved to curvenote and journal API later - like checking authorisation, checking for public versus private buckets based on the venue or domains settings, and authorisations on the curvenote account.

- deployment should default to private bucket/cdn if not configured otherwise
- can still deploy to public if domains are specified
- can deploy to private bucket and submit to a venue

New cli option is `--venue <name>`

new behaviour is:

 * if `--venue` upload privately, submit to venue
 * if `config.site.domains` or `--domain` upload publicly, deploy to domains
 * else upload privately and return cdnkey

## UPDATE

we are now aiming for a different CLI UX https://curvenote.com/oxa:ldzjE1Zd91VYtnu0M52O/e6L79XEO7YzutWpnbw4n.1


packages/curvenote/src/web/deploy.ts
```diff
@@ -112,20 +114,25 @@ export async function deployContentToCdn(session: ISession, opts?: { ci?: boolea
...
   };
-  const { json: uploadTargets } = await session.post<SiteUploadResponse>(
-    '/sites/upload',
-    uploadRequest,
-  );
 
+  return { files, uploadRequest };
+}
+
+export async function processUpload(
```
fwkoch:
```suggestion
export async function performFileUploads(
```
or something? Just to be slightly more explicit.

packages/curvenote/src/web/deploy.ts
```diff
@@ -151,23 +158,129 @@ export async function deployContentToCdn(session: ISession, opts?: { ci?: boolea
...
+  cdnKey: string,
+  venue: string,
+  username: string,
+) {
+  const toc = tic();
+  const submissionRequest: VenueSubmitRequest = {
+    id: cdnKey,
+    username,
+  };
+  const deployResp = { ok: true, submissionRequest }; //await session.post(`/venues/${venue}/submit`, deployRequest);
```
fwkoch: :+1: just a placeholder

packages/curvenote/src/web/deploy.ts
```diff
@@ -211,36 +328,126 @@ export async function promoteContent(session: ISession, cdnKey: string, domains?
...
+    // TODO public-venue?
+    case 'public': {
+      if (!domains || domains.length === 0) {
+        throw new Error(`🚨 Internal Error: No domains specified during public deployment`);
+      }
+      await confirmOrExit(
+        `Deploy local content to "${domains.map((d) => `https://${d}`).join('", "')}"?`,
+        opts,
+      );
+      await preflightPromotePublicContent(session, domains); // TODO check domains exist, and user can promote to them
```
fwkoch: Ah, nice, so these preflights are for early reporting in the CLI (i.e. no need to wait for failures). They are not necessary for the actual promotion.

packages/curvenote/src/web/deploy.ts
```diff
@@ -151,23 +158,129 @@ export async function deployContentToCdn(session: ISession, opts?: { ci?: boolea
...
+}
+
+/**
+ * Deploy content to the public/private CDN based on the usePublicCdn argument
+ *
+ * @param session
+ * @param usePublicCdn whether to deploy to the public CDN or default private CDN
+ * @param cdnKey
+ * @param filepaths
+ * @returns cdnkey
```
fwkoch: This `cdnKey` is just the same as the `cdnKey` input. Doesn't change when content is deployed.

fwkoch (approved):
This looks great - nice placeholders for submission API. It took me a little bit to remember what each step does (upload, deploy, promote) and how the CDN key comes into play (it's just an id for the content...). But all of this complexity is nicely contained to the code and not exposed to the CLI.

stevejpurves:
ok some changes to allow our containers to still work as previously. We need to merge this and release it **before** we can update our api!
cc @fwkoch

stevejpurves:
Note this PR is currently built against https://github.com/executablebooks/mystmd/pull/587

packages/curvenote/src/works/create.ts
```diff
@@ -0,0 +1,55 @@
...
+      '⚠️ You must be authenticated for this call. Use `curvenote token set [token]`',
+    );
+  }
+
+  const siteConfig = selectors.selectCurrentSiteConfig(session.store.getState());
+  if (!siteConfig) {
+    throw new Error('🧐 No site config found.');
+  }
+
+  const transferData = await loadTransferFile();
```
stevejpurves: @fwkoch can you point me to a better way to be interacting with this file. Right now i'm reading/writing in the current folder but really/probably this file should be created and maintained adjacent to the root curvenote/myst.yml file?

I'm not sure how to achieve that, and double unsure of how to do that without involving any changes to mystmd.

packages/curvenote/src/session/session.ts
```diff
@@ -71,6 +78,7 @@ export class Session implements ISession {
     if (this.SITE_URL !== DEFAULT_SITE_URL) {
       this.log.warn(`Connecting to Site at: "${this.SITE_URL}".`);
     }
+    this.JOURNALS_URL = STAGING_JOURNALS_API_URL;
```
stevejpurves: We are hard coding to the staging environment for now (until we get latest journals API into prod)

In order to properly be able to test in staging, we are going to need to bring up a staging CDN mirroring prod -- which we have not done yet. Then we can select the URl base don the token as we do with the SITE_URL. cc @rowanc1 

Otherwise we need a mechanisms to set the api and other URLS. It seems like `SessionOptions` provides for this but I cannot see anywhere were we can pass those in, `--apiUrl` is not an argument on the cli yet -- @fwkoch was that the intention? this is a bit cumbersome for testing too, should we do something via environment variable instead allowing it to be set in the background? this would make both local and staging testing easier.

packages/curvenote/src/web/deploy.ts
stevejpurves: should be dealt with now

packages/curvenote/src/web/deploy.ts
stevejpurves: yes but they are placeholders at the moment! we need a `check` api call or something to enable these efficiently

PR #446, opened 2023-10-18
align init with mystmd

stevejpurves:
* init from folder creates a flat site nav
* write-toc works on both fresh init and init on existing project

### DIFF
--- a/packages/curvenote/src/sync/init.ts
+++ b/packages/curvenote/src/sync/init.ts
@@ -9,6 +9,9 @@ import {
   selectors,
   writeConfigs,
   startServer,
+  writeTocFromProject,
+  validateTOC,
+  projectFromPath,
 } from 'myst-cli';
 import { LogLevel, writeFileToFolder } from 'myst-cli-utils';
 import type { ProjectConfig } from 'myst-config';
@@ -83,26 +86,26 @@ ${docLinks.overview}
 export async function init(session: ISession, opts: Options) {
   if (!opts.yes) session.log.info(await WELCOME(session));
   if (opts.domain) session.log.info(`Using custom domain ${opts.domain}`);
-  let path = resolve('.');
+  let currentPath = resolve('.');
   // Initialize config - error if it exists
-  if (selectors.selectLocalSiteConfig(session.store.getState(), path)) {
+  if (selectors.selectLocalSiteConfig(session.store.getState(), currentPath) && !opts.writeToc) {
     throw Error(
       `Site config in ${CURVENOTE_YML} config already exists, did you mean to ${chalk.bold(
         'curvenote clone',
       )} or ${chalk.bold('curvenote start')}?`,
     );
   }
-  const folderName = basename(path);
+  const folderName = basename(currentPath);
   const siteConfig = getDefaultSiteConfig(folderName);
 
   // Load the user now, and wait for it below!
   let me: MyUser | Promise<MyUser> | undefined;
   if (!session.isAnon) me = new MyUser(session).get();
 
-  const folderIsEmpty = fs.readdirSync(path).length === 0;
+  const folderIsEmpty = fs.readdirSync(currentPath).length === 0;
   if (folderIsEmpty && opts.yes) throw Error('Cannot initialize an empty folder');
   let content;
-  const projectConfigPaths = findProjectsOnPath(session, path);
+  const projectConfigPaths = findProjectsOnPath(session, currentPath);
   if ((!folderIsEmpty && opts.yes) || projectConfigPaths.length) {
     content = 'folder';
   } else {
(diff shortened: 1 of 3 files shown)


packages/curvenote/src/sync/init.ts
```diff
@@ -115,48 +118,66 @@ export async function init(session: ISession, opts: Options) {
...
+          if (!projectConfigPaths.length) {
+            throw Error(`No markdown or notebook files found`);
+          }
+          session.log.info(`🧹 No additional markdown or notebook files found`);
         }
-        session.log.info(`🧹 No additional markdown or notebook files found`);
       }
     }
-    siteConfig.projects = projectConfigPaths.map((p) => ({ path: p, slug: basename(resolve(p)) }));
+    siteConfig.projects = undefined;
```
fwkoch: This is aligning with MyST which does not automatically populate projects in the site config 👍 

Should we just update `getDefaultSiteConfig` to not set `projects = []` by default, i.e. remove this line: https://github.com/curvenote/curvenote/blob/b122528e0186630d130b59a0b8f0e900745f63d6/packages/curvenote/src/sync/utils.ts#L24

Then we don't have to explicitly set to `undefined`

packages/curvenote/src/sync/init.ts
```diff
@@ -115,48 +118,66 @@ export async function init(session: ISession, opts: Options) {
...
+        return;
+      } else {
+        const project = projectFromPath(session, currentPath);
+        session.log.info(
+          `📓 Writing '_toc.yml' file to ${
+            currentPath === '.' ? 'the current directory' : currentPath
+          }`,
+        );
+        if (!project) throw Error(`Could not load project from ${currentPath}`);
+        writeTocFromProject(project, currentPath);
```
fwkoch, line 144: This is writing `_toc` to the implicit project from the current path, rather than the loaded `projectConfigPaths` - these usually match, but not necessarily. Maybe we should write toc to those paths instead?

packages/curvenote/src/sync/init.ts
```diff
@@ -83,26 +86,26 @@ ${docLinks.overview}
 export async function init(session: ISession, opts: Options) {
   if (!opts.yes) session.log.info(await WELCOME(session));
   if (opts.domain) session.log.info(`Using custom domain ${opts.domain}`);
-  let path = resolve('.');
+  let currentPath = resolve('.');
   // Initialize config - error if it exists
-  if (selectors.selectLocalSiteConfig(session.store.getState(), path)) {
+  if (selectors.selectLocalSiteConfig(session.store.getState(), currentPath) && !opts.writeToc) {
```
fwkoch, line 91: This feels not quite right? If I have already created a site config but want to initialize a `_toc` file, calling `curvenote init --write-toc` would allow this function to continue, resetting the site config to default before writing the toc.

Should we have an alternative path to only write toc in the case of existing site config?

fwkoch (commented):
A few things about this seem a little off - pretty sure it will work fine in normal cases where (1) you have not created a config file yet and (2) project and site are co-located. But I think if you diverge from that, behaviour may be unexpected...

packages/curvenote/src/sync/init.ts
fwkoch: Maybe I'm wrong about this? Testing some, the behaviour seems ok - doesn't match what I expected from looking at the code.

fwkoch: Oh, sheesh, I totally missed the early `return`.

packages/curvenote/src/sync/init.ts
fwkoch: Ok - confirmed that this does happen. If `site` config exists looking something like this:
```yaml
...
site:
  projects:
    - path: subproject
      slug: sub
```
It will still just try to write the `_toc` in the current folder, as opposed to `subproject` folder. That's a bit of an edge case though, I think it's fine to leave as is.

packages/curvenote/src/sync/init.ts
```diff
@@ -115,48 +118,66 @@ export async function init(session: ISession, opts: Options) {
   let pullComplete = false;
   let title = projectConfig?.title || siteConfig.title || undefined;
   if (content === 'folder') {
-    if (projectConfigPaths.length) {
+    if (projectConfigPaths.length > 0) {
       const pathListString = projectConfigPaths
         .map((p) => `  - ${join(p, CURVENOTE_YML)}`)
```
fwkoch, line 123: At some point we could pull some updated functionality from myst, for example, like to get the correct existing yaml file name here, rather than just printing `curvenote.yml` always. Certainly out of scope for this PR - just log messaging stuff.

packages/curvenote/src/sync/init.ts
stevejpurves: ok - out of curiosity though, the change here would be to write toc where? to `projectConfigPaths[0]`? or individually write a toc to each [sub]path? e.g. `projectConfigPaths.forEach((p) =>  writeTocFromProject(the-right-project-somehow, p)`

stevejpurves: happy to take this PR as is though as it meets the initial local use case we're trying to cover just now

PR #447, opened 2023-10-19
🧹👷🏽 clean rebuild on `works create`

stevejpurves:
this ensures that assets such as `meca.zip` and other things that might be needed online for the work and submission are [re]built and available.


stevejpurves:
@fwkoch bringing this one in quickly -- perhaps there was a better way to reuse mystmd functionality here? comments welcome and we can make changes in another PR.

PR #451, opened 2023-10-27
✅ Improve checks

rowanc1:


packages/curvenote/src/cli/check.ts
```diff
@@ -2,21 +2,21 @@ import { Command } from 'commander';
...
     [
-      { id: 'abstract-exists' },
-      { id: 'abstract-length', max: '400' },
-      { id: 'availability-exists' },
-      { id: 'links-resolve' },
+      // { id: 'abstract-exists' },
+      // { id: 'abstract-length', max: '400' },
+      // { id: 'availability-exists' },
+      // { id: 'links-resolve' },
+      ...submissionRuleChecks,
```
fwkoch: Could add this, just to visually differentiate between checks and check definitions... Doesn't actually change anything:

```suggestion
      ...submissionRuleChecks.map(({ id }) => {
        return { id };
      }),
```

(Also no need for the comments, I don't think.)

fwkoch (approved):
Nice - pretty straightforward relocating a bunch of code.

rowanc1:
Nice, thank you!

PR #455, opened 2023-11-06
🚓 pre-flight style check for user submissions

stevejpurves:
Calls the journals API to check whether a user can submit to a particular venue


rowanc1:
Nice!

PR #461, opened 2024-01-10
flat projects like mystmd on init from curvenote

stevejpurves:
previously running `init` on a local folder behaved like `mystmd`, with this PR we get the same behavior on `init` from a remote curvenote project. `pull` and `clone` continue to work as expected with the new layout.

### DIFF
--- a/packages/curvenote/src/sync/clone.ts
+++ b/packages/curvenote/src/sync/clone.ts
@@ -1,7 +1,6 @@
 import chalk from 'chalk';
 import fs from 'node:fs';
 import inquirer from 'inquirer';
-import { join } from 'node:path';
 import { loadConfig, selectors, writeConfigs } from 'myst-cli';
 import { LogLevel } from 'myst-cli-utils';
 import type { ProjectConfig, SiteConfig } from 'myst-config';
@@ -39,15 +38,14 @@ export async function interactiveCloneQuestions(
     }
   }
   let path: string;
-  const defaultPath = join('content', project.data.name);
+  const defaultPath = '.';
   if (opts?.path || opts?.yes) {
     path = opts?.path ?? defaultPath;
     if (path !== '.' && fs.existsSync(path)) {
       throw new Error(`Invalid path for clone: "${path}", it must not exist.`);
     }
   } else {
-    const { projectPath } = await inquirer.prompt([questions.projectPath(defaultPath)]);
-    path = projectPath;
+    path = defaultPath;
   }
   try {
     // Throw if project doesn't exist - that's what we want!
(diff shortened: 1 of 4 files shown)


packages/curvenote/src/sync/clone.ts
```diff
@@ -39,26 +38,29 @@ export async function interactiveCloneQuestions(
...
   } catch {
     // Project config does not exist; good!
-    // TODO: Add all sorts of other stuff for the project data that we know!!
-    const projectConfig = getDefaultProjectConfig(project.data.title);
-    projectConfig.remote = project.data.id;
-    projectConfig.description = project.data.description;
+    const nonNullProjectData = Object.fromEntries(
+      Object.entries(project.data).map(([key, item]) => [key, item === null ? undefined : item]),
+    );
+    const projectConfig = {
```
fwkoch, line 59: This is a nice improvement.

packages/curvenote/src/sync/init.ts
```diff
@@ -170,8 +170,6 @@ export async function init(session: ISession, opts: Options) {
     projectConfig = results.projectConfig;
     title = projectConfig.title;
     currentPath = siteProject.path;
-    siteConfig.projects = [siteProject];
```
fwkoch, line 173: Quietly deprecating multi-project behaviour 👍

packages/curvenote/src/sync/clone.ts
```diff
@@ -39,26 +38,29 @@ export async function interactiveCloneQuestions(
     }
   }
   let path: string;
-  const defaultPath = join('content', project.data.name);
+  const defaultPath = '.';
```
stevejpurves, line 41: So in curvenote's launchpad-server, path comes into this function as `undefined` which means the default path is applied, this change also modifies the behaviour of the client library based usage (e.g. `sync.clone`) so that with no change to the calling code, we also get the same "flat project" behaviour as with `curvenote init`. That seems good!?

fwkoch, line 41: This seems fine - is there a worry we clobber existing stuff in the current folder though?

packages/curvenote/src/sync/clone.ts
```diff
@@ -39,26 +38,29 @@ export async function interactiveCloneQuestions(
...
   if (opts?.path || opts?.yes) {
     path = opts?.path ?? defaultPath;
     if (path !== '.' && fs.existsSync(path)) {
       throw new Error(`Invalid path for clone: "${path}", it must not exist.`);
     }
   } else {
-    const { projectPath } = await inquirer.prompt([questions.projectPath(defaultPath)]);
-    path = projectPath;
+    path = defaultPath;
```
fwkoch, line 48: `path` above can still be set as a CLI option, so now it's just no longer an interactive choice.

fwkoch (approved):

packages/curvenote/src/sync/clone.ts
stevejpurves: we're ok for now I think, but otherwise would want to putt a warning/prompt in though

PR #481, opened 2024-02-16
♻️ build reports on all submissions, some uniformizing of log data

stevejpurves:
build pages for everyone

### DIFF
--- a/packages/curvenote/src/submissions/submit.utils.ts
+++ b/packages/curvenote/src/submissions/submit.utils.ts
@@ -240,6 +240,7 @@ export async function confirmUpdateToExistingSubmission(
 
 export async function createNewSubmission(
   session: ISession,
+  logCollector: Record<string, any>,
   venue: string,
   kind: string,
   cdnKey: string,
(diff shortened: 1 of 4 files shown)


packages/curvenote/src/submissions/submit.utils.ts
```diff
@@ -236,6 +238,41 @@ export async function confirmUpdateToExistingSubmission(
...
+  if (opts?.draft) {
+    session.log.info(`🚀 ${chalk.green(`Your draft was successfully submitted to "${venue}"`)}.`);
+  } else {
+    session.log.info(`🚀 ${chalk.green(`Your work was successfully submitted to "${venue}"`)}.`);
+  }
+
+  logCollector.work = work;
+  logCollector.workVersion = workVersion;
+  logCollector.submission = submission;
+  logCollector.submissionVersion = submissionVersion;
```
stevejpurves, line 273: now logging the exact same info here that goes into the `transfer.yml`, that means `work.id` etc... rather than `workId`
@rowanc1 not sure if that means you need to make a tweak in the action?

rowanc1: No, haven't started using this yet!

PR #486, opened 2024-02-20
🔑 key based submissions

stevejpurves:
* can submit with --key option
* we sniff the git repo details for the build/job source info

### DIFF
--- a/packages/curvenote-cli/src/submissions/submit.utils.ts
+++ b/packages/curvenote-cli/src/submissions/submit.utils.ts
@@ -20,16 +20,14 @@ import {
   postUpdateSubmissionWorkVersion,
 } from './utils.js';
 import inquirer from 'inquirer';
+import type { SubmissionDTO } from '@curvenote/common';
 
 export type SubmitOpts = {
   kind?: string;
   yes?: boolean;
   info: boolean;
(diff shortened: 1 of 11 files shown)


packages/curvenote/package.json
```diff
@@ -42,13 +42,14 @@
     "build": "npm-run-all -l clean copy:version -p build:cli"
   },
   "dependencies": {
+    "isomorphic-git": "^1.25.6",
```
rowanc1: This should be put in dev-deps.

packages/curvenote-cli/src/submissions/submit.utils.ts
```diff
@@ -169,6 +167,21 @@ export async function checkVenueAccess(session: ISession, venue: string) {
   }
 }
 
+export async function checkForSubmissionUsingKey(session: ISession, venue: string, key: string) {
+  session.log.debug(`checking for existing submission using key "${key}"`);
+  try {
+    const submission = await getFromJournals(session, `sites/${venue}/submissions/by?key=${key}`);
```
rowanc1: @stevejpurves we talked about maybe changing this to `/sites/${venue}/submissions?key=${key}`?

PR #492, opened 2024-02-23
📤 Update the export and clean cli

fwkoch:
This adds `typst`, `jats`, and `meca` commands to the `curvenote export` CLI. It also adds all of these as well as `--cache` to the `curvenote clean` CLI. All of this follows the exact same pattern as existing work. It is dependent on https://github.com/executablebooks/mystmd/pull/935

### DIFF
--- a/packages/curvenote/src/web.ts
+++ b/packages/curvenote/src/web.ts
@@ -9,10 +9,6 @@ import {
   makeCheckLinksOption,
   makeKeepHostOption,
   makeHeadlessOption,
-  makePdfOption,
-  makeTexOption,
-  makeDocxOption,
-  makeSiteOption,
   makeDomainOption,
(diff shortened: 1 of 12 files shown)


packages/curvenote/src/web.ts
```diff
@@ -30,14 +26,7 @@ function makeCurvenoteStartCLI(program: Command) {
 
 function makeBuildCLI(program: Command) {
   const command = new Command('build')
-    .description(
-      'Build pdf, tex, and word exports from MyST files as well as build MyST site content',
-    )
-    .argument('[files...]', 'list of files to export')
-    .addOption(makePdfOption('Build'))
```
fwkoch, line 37: None of this stuff worked anyway... the function we invoke from this command only does `site` build, no matter what.

PR #493, opened 2024-02-23
🌎 Consume session fetch and https proxy updates from myst

fwkoch:
Consumes the changes to the `ISession` interface here: https://github.com/executablebooks/mystmd/pull/933


stevejpurves:
when testing this out validating the token vs our api fails for some reason, debugging that locally...

PR #497, opened 2024-02-28
✨ use new upload API

stevejpurves:
- [x] refactoring
- [x] use for submissions
- [x] use for web deploys
- [x] resume/retry failed uploads


packages/curvenote-cli/src/session/session.ts
```diff
@@ -83,14 +84,17 @@ export class Session implements ISession {
...
     if (url?.startsWith(STAGING_API_URL)) {
       this.JOURNALS_URL = STAGING_SITES_API_URL;
       this.PRIVATE_CDN = 'https://prv.curvenote.dev';
+      this.TEMP_CDN = 'https://tmp.curvenote.dev';
       this.PUBLIC_CDN = 'https://cdn.curvenote.dev';
     } else if (url?.startsWith(LOCAL_API_URL)) {
       this.JOURNALS_URL = LOCAL_SITES_API_URL;
       this.PRIVATE_CDN = 'https://prv.curvenote.dev';
+      this.TEMP_CDN = 'https://tmp.curvenote.dev';
       this.PUBLIC_CDN = 'https://cdn.curvenote.dev';
```
stevejpurves, line 99: soon the cdn specs will be recovered from sites API, based on submission venue, kind & workflow etc...

packages/curvenote-cli/src/uploads/resumable.utils.ts
```diff
@@ -0,0 +1,121 @@
...
+      headers['Content-Range'] = contentRange;
+    }
+    return session.fetch(location, {
+      method: 'PUT',
+      headers,
+      body: readStream,
+    });
+  } catch (e) {
+    // eat error
+  }
```
stevejpurves, line 46: recover from failures as we want to retry/resume

packages/curvenote-cli/src/uploads/resumable.utils.ts
```diff
@@ -0,0 +1,121 @@
...
+      'x-goog-resumable': 'start',
+      'content-type': upload.contentType,
+    },
+  });
+
+  if (!resumableSession.ok) {
+    session.log.error(`Failed to start upload for ${upload.from}`);
+    session.log.error(`${resumableSession.status} ${resumableSession.statusText}`);
+    throw new Error(`Failed to start upload for ${upload.from}`);
+  }
```
stevejpurves, line 84: we have always been initiating a resumable session, so nothing new above.

packages/curvenote-cli/src/uploads/resumable.utils.ts
```diff
@@ -0,0 +1,121 @@
...
+  // Endpoint to which we should upload the file
+  const location = resumableSession.headers.get('location') as string;
+
+  let retries = 0;
+  const numberOfRetries = 3;
+  const numberOfResumes = 10;
+  for (; retries < numberOfRetries; retries++) {
+    await doTheUpload(session, location, upload);
+    const initialCheckResponse = await checkUploadCompleted(session, location);
```
stevejpurves, line 93: we attempt the upload, and make a separate call to check whether it completed

packages/curvenote-cli/src/uploads/resumable.utils.ts
```diff
@@ -0,0 +1,121 @@
...
+    if (initialCheckResponse.complete) {
+      break;
+    }
+    if (initialCheckResponse.range && opts?.resume) {
```
stevejpurves, line 97: if we have opted in to resuming, this code will attempt to resume -- only really useful on larger uploads. Also I tried to test this code by aborting the readstream after 150ms but could not get a realistic upload failure, aborting the stream caused an exception in node. So i've put this behind an option until we test it more, if we use the --resume flag on our submit runs, we'll encounter realistic failures.

reference: https://cloud.google.com/storage/docs/performing-resumable-uploads

packages/curvenote-cli/src/uploads/resumable.utils.ts
stevejpurves: no matter what, we'll attempt 3 retries of the upload from scratch

PR #508, opened 2024-03-14
publish with key

stevejpurves:
Update the `curvenote submissions publish/unpublish` to work with the `--key`


stevejpurves:
I think @rowanc1 needs this too?
cc @fwkoch 

@rowanc1 I am assuming that you are manually patching through the myst.yml#project.id into the `--key`?

PR #511, opened 2024-03-20
🗂 submit to collections

stevejpurves:
reworking of the submission process to work with collections, including a new `--collection` cli option.

### DIFF
--- a/packages/curvenote-cli/src/submissions/submit.utils.ts
+++ b/packages/curvenote-cli/src/submissions/submit.utils.ts
@@ -20,10 +20,18 @@ import {
   postUpdateSubmissionWorkVersion,
 } from './utils.js';
 import inquirer from 'inquirer';
-import type { SubmissionsListItemDTO, SubmissionsListingDTO } from '@curvenote/common';
+import type {
+  CollectionDTO,
+  CollectionsDTO,
+  SubmissionDTO,
+  SubmissionKindDTO,
+  SubmissionsListItemDTO,
+  SubmissionsListingDTO,
+} from '@curvenote/common';
 
 export type SubmitOpts = {
   kind?: string;
+  collection?: string;
   yes?: boolean;
   info: boolean;
   draft?: boolean;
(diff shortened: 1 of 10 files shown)


packages/curvenote-cli/src/submissions/submit.utils.ts
```diff
@@ -61,41 +87,123 @@ export function venueQuestion(session: ISession) {
...
+  let kindId;
   if (opts?.kind) {
-    const match = kinds.items.find(
+    const match = kinds.find(
       ({ name }: { name: string }) => name.toLowerCase() === opts.kind?.toLowerCase(),
     );
     if (!match) {
       session.log.info(
-        `${chalk.red(`🚨 submission kind "${opts.kind}" is not accepted at venue ${venue}`)}`,
+        `${chalk.bold.red(`⛔️ submission kind "${opts.kind}" is not accepted in the collection "${collectionMoniker(collection)}"`)}`,
```
fwkoch, line 185: At some point we will probably want to massage this logging so it covers the UX for both (1) site with collections and (2) site with only a default collection but no collections as far as the user is concerned.

I think less nuance is ok for now though...

packages/curvenote-cli/src/submissions/submit.utils.ts
```diff
@@ -61,41 +87,123 @@ export function venueQuestion(session: ISession) {
...
-  } catch (err: any) {
+  const openCollections = collections.items.filter((c) => c.open);
+  if (openCollections.length === 0) {
+    session.log.info(
+      `${chalk.red(`⛔️ no collections are open for submissions at venue "${venue}"`)}`,
+    );
+    process.exit(1);
+  }
+
+  session.log.debug(`Explicit collection provided: ${opts?.collection}`);
```
fwkoch: Minor stuff, but this should probably only be logged this way if `opts.collection` is defined.

fwkoch (approved):
Looking good - my only comments are pretty trivial logging stuff.

I'll put this through the wringer a bit more as I combine it with my work removing `transfer.yml` in favour of `project.id`.

packages/curvenote-cli/src/submissions/submit.utils.ts
stevejpurves: yeah the messaging UX overall probably could be improved

PR #518, opened 2024-04-04
🔧 Fix submitting after only draft submissions

fwkoch:
When creating a real submission after only making drafts, the submission would fail due to flaws in existing submission checks. This PR does two things:

1. It fixes the existing submission check logic, so now drafts are correctly ignored on submission creation.

~2. Drafts also submit with `key: undefined` - this creates a new _work_ in addition to a new draft submission.~ This change was reverted; drafts remain tied to the same work!


packages/curvenote-cli/src/submissions/submit.ts
```diff
@@ -71,15 +72,15 @@ export async function submit(session: ISession, venue: string, opts?: SubmitOpts
     path: gitInfo?.path,
     commit: gitInfo?.commit,
   };
-
-  session.log.info(`📍 Submitting using key: ${chalk.bold(key)}`);
+  if (opts?.draft) key = undefined;
```
fwkoch: @stevejpurves - Once we have a better idea of workflows we can probably delete this (could even delete now...). Things are fixed so even without this line, draft submission workflow is ok.

The difference is: with this line, drafts create new "inaccessible" work with no key and a new submission version; without this line, drafts only create work _versions_ on the existing keyed work (still a new submission version).

PR #524, opened 2024-04-06
⛔️ Do not prompt on --yes option, fail instead

fwkoch:
Even when providing CLI flag `-y`, user was still prompted for `venue`, `collection`, and `kind`.

Now, the CLI will hard fail if `-y` is used but the submission cannot be completed due to ambiguity of these fields.


fwkoch:
This is related to the updated action here: https://github.com/curvenote/actions/pull/5

In that case, we want to fail fast with a useful error message if `collection` is not provided.

PR #527, opened 2024-04-08
🥇 Prioritize default collections on selection

fwkoch:
This updates the collection selection CLI:

- if `--collection` is provided, no changes, only better error messaging
- if `-y` is provided:
  - Open default collection will be selected, even if there are are multiple open collections
  - Closed default collection will error (even if there are open collections)
- if `-y` is not provided:
  - Open default collections will be at the top of the selection list
  - User will be warned if default collection is closed.

### DIFF
--- a/packages/curvenote-cli/src/submissions/submit.utils.ts
+++ b/packages/curvenote-cli/src/submissions/submit.utils.ts
@@ -24,7 +24,6 @@ import inquirer from 'inquirer';
 import type {
   CollectionDTO,
   CollectionListingDTO,
-  SubmissionDTO,
   SubmissionKindDTO,
   SubmissionKindListingDTO,
   SubmissionsListItemDTO,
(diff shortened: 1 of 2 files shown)


packages/curvenote-cli/src/submissions/submit.utils.ts
```diff
@@ -85,7 +84,10 @@ export async function determineCollectionAndKind(
   collections: CollectionListingDTO,
   opts?: { kind?: string; collection?: string; yes?: boolean },
 ) {
-  const openCollections = collections.items.filter((c) => c.open);
+  const openCollections = [
+    ...collections.items.filter((c) => c.open && c.default),
```
fwkoch, line 88: Default at the top - we could also make it bold or `... (default)` or something in the selection choices...

packages/curvenote-cli/src/submissions/submit.utils.ts
```diff
@@ -95,7 +97,7 @@ export async function determineCollectionAndKind(
 
   if (opts?.collection) session.log.debug(`Explicit collection provided: ${opts?.collection}`);
   let selectedCollection = opts?.collection
-    ? openCollections.find((c) => c.name === opts?.collection)
+    ? collections.items.find((c) => c.name === opts?.collection)
```
fwkoch, line 100: This fixes up the error messaging so user gets `selection is closed` error rather than `selection does not exist`.

PR #528, opened 2024-04-09
🐛 Fix 'Bearer undefined' bug

fwkoch:
Anonymous sessions should not have a `Bearer` token - we were accidentally setting `Bearer undefined` - just a simple bug.

I also added a command `curvenote token anon` so we can run curvenote unauthenticated without deleting the saved tokens.

### DIFF
--- a/packages/curvenote-cli/src/session/session.ts
+++ b/packages/curvenote-cli/src/session/session.ts
@@ -87,12 +87,12 @@ export class Session implements ISession {
     this.PRIVATE_CDN = 'https://prv.curvenote.com';
     this.TEMP_CDN = 'https://tmp.curvenote.com';
     this.PUBLIC_CDN = 'https://cdn.curvenote.com';
-    if (url?.startsWith(STAGING_API_URL)) {
+    if (this.API_URL?.startsWith(STAGING_API_URL)) {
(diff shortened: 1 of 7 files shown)


packages/curvenote-cli/src/session/session.ts
```diff
@@ -87,12 +87,12 @@ export class Session implements ISession {
     this.PRIVATE_CDN = 'https://prv.curvenote.com';
     this.TEMP_CDN = 'https://tmp.curvenote.com';
     this.PUBLIC_CDN = 'https://cdn.curvenote.com';
-    if (url?.startsWith(STAGING_API_URL)) {
+    if (this.API_URL?.startsWith(STAGING_API_URL)) {
```
fwkoch, line 90: @stevejpurves - is this change ok? It seems like we should make this selection based on the resolved `API_URL` not the token's `url`.

packages/curvenote-cli/src/session/tokens.ts
```diff
@@ -80,7 +80,7 @@ export async function getHeaders(
     'X-Client-Version': CLIENT_VERSION,
   };
   const sessionToken = await getSessionToken(session, tokens);
-  if (session) {
+  if (sessionToken) {
```
fwkoch, line 83: 🚨 This is the actual fix

PR #531, opened 2024-04-10
🧾 custom listing directive

stevejpurves:
Adding custom directives to curvenote cli

### DIFF
--- a/packages/curvenote-cli/src/directives/listing.ts
+++ b/packages/curvenote-cli/src/directives/listing.ts
@@ -1,0 +1,71 @@
+import type { DirectiveData } from 'myst-common';
+import type { VFile } from 'vfile';
+
+export const listingDirective = {
+  name: 'listing',
+  doc: 'A listing directive that can be used to show a list of articles from a specific venue, collection or kind.',
+  arg: {
+    type: 'myst',
+    doc: 'The title of the listing.',
+  },
+  body: {
+    type: 'myst',
+    doc: 'Descriptive content to be displayed along with the listing.',
+  },
+  options: {
+    venue: {
+      type: 'string',
+      doc: 'The venue to list articles from.',
+      required: true,
+    },
+    collection: {
+      type: 'string',
+      doc: 'The collection to list articles from.',
+      required: false,
+    },
+    status: {
+      type: 'string',
+      doc: 'The status of articles to list (published | in-review).',
+      required: false,
+    },
+    limit: {
+      type: 'number',
+      doc: 'The maximum number of articles to list.',
+      required: false,
+    },
+    kind: {
+      type: 'string',
+      doc: 'The kind of listing to display (list | cards).',
+      required: false,
+    },
+  },
+  validate(data: DirectiveData, vfile: VFile) {
+    if (!data.options?.venue) {
+      vfile.message('A venue must be supplied.');
+    }
+    if (
+      data.options?.status &&
+      (typeof data.options.status !== 'string' ||
+        (typeof data.options.status === 'string' &&
+          !['published', 'in-review'].includes(data.options.status)))
+    ) {
+      vfile.message('Invalid status supplied.');
+    }
+
+    // TODO lookup API and validate venue, collection, kind exist?
+    // TODO how to validate against the correct API? dev/staging/prod?
+
+    return data;
+  },
+  run(data: DirectiveData) {
+    return [
+      {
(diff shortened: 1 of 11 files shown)


stevejpurves:
@fwkoch In terms of being able to register additional directives directly, does thats mean being able to inject them down to this level? https://github.com/executablebooks/mystmd/blob/3c04fd38d8574aa3473cc36acdab37982097b895/packages/myst-cli/src/process/mdast.ts#L96

packages/curvenote-cli/src/directives/listing.ts
```diff
@@ -0,0 +1,71 @@
+import type { DirectiveData } from 'myst-common';
+import type { VFile } from 'vfile';
+
+export const listingDirective = {
+  name: 'listing',
```
rowanc1: Maybe `list:articles` as the type? Thinking we might want other listings in the future. Or `cn:articles`? As these are curvenote directives.

packages/curvenote-cli/src/directives/listing.ts
```diff
@@ -0,0 +1,71 @@
...
+      type: 'string',
+      doc: 'The status of articles to list (published | in-review).',
+      required: false,
+    },
+    limit: {
+      type: 'number',
+      doc: 'The maximum number of articles to list.',
+      required: false,
+    },
+    kind: {
```
rowanc1: We have usually reserved kind for a sub-type (admonition > warning, proof > example, container > figure). This could be `display: cards` or `class: cards`

fwkoch:
Directives get injected here: https://github.com/executablebooks/mystmd/blob/8ca2c518923ac8f2a64c6054adb58b3303bd5518/packages/myst-cli/src/process/myst.ts#L40-L48 (this is at the file-parse level, before the process step you linked in your comment above @stevejpurves )

Currently, there are (1) built-in myst directives and (2) plugin directives. We could get this working immediately by just adding curvenote directives as plugins that are _always_ loaded on a curvenote session. That could come in here: https://github.com/curvenote/curvenote/blob/29f52dcd58476b82ff4dceb7252651776ea87101/packages/curvenote-cli/src/session/session.ts#L198-L204 - after the call to `loadCurvenotePlugins`, just append any additional curvenote directives to the list. Another option (3) is that we extend `parseMyst` to take `directives` as another `opt`... that would take a lot of finagling to expose all the way up to curvenote though, and would be easy to mess up since `parseMyst` is called in several places...

I like the "implicit curvenote plugin" way of doing things...

stevejpurves:
Thanks for the feedback @fwkoch @rowanc1, directive now being loaded as a plugin. When it comes time to be independently shipping this, probably `src/plugins` becomes it's own package...

packages/curvenote-cli/src/plugins/articles.ts
```diff
@@ -0,0 +1,75 @@
...
+      required: false,
+    },
+    limit: {
+      type: Number,
+      doc: 'The maximum number of articles to list.',
+      required: false,
+    },
+    pagination: {
+      type: String,
+      doc: 'If `limit` is set, this gives a hint to the renderer on how to handle pagination (more | all | infinite).',
```
fwkoch: 🤔 We could extend these directive spec options to use the same `options` validation we are using for template / check options... that would give us access to more declarative validation like `choices`...

Just musing - no influence on this PR 🙂

packages/curvenote-cli/src/plugins/directives/articles.ts
```diff
@@ -0,0 +1,74 @@
...
+  validate(data, vfile) {
+    if (!data.options?.venue) {
+      vfile.message('A venue must be supplied.');
+    }
+
+    validateStringOptions(vfile, 'status', data.options?.status, ['published', 'in-review']);
+    validateStringOptions(vfile, 'layout', data.options?.layout, ['list', 'cards']);
+    validateStringOptions(vfile, 'pagination', data.options?.pagination, ['more', 'all', 'scroll']);
+
+    // TODO get hold of session?
```
stevejpurves, line 58: @fwkoch how could i get hold of the session here?

fwkoch: Great question - the current interface does not allow that, which is a big limitation on the capabilities of these plugins. I think it would be super reasonable to pass the session in and allow these plugins to do more complicated things than just manipulate nodes.

Issue #539, opened 2024-04-16
`curvenote check` does not allow checking on any kind, or collection


stevejpurves:
Currently the CLI for checks i limited to:

```
% curvenote check scipy --help
Usage: curvenote check [options] [venue]

Run checks from plugins or example checks on your MyST project

Arguments:
  venue            Venue to check the submission against locally

Options:
  --kind <string>  Submit to the venue using this submission kind
  -h, --help       display help for command
```

and when you run `curvenote check` only the default submission kind in checked.

## Proposal
This should be extended to accommodate multiple kinds and collections. i.e.

when running `curvenote check venue`
1. in a site where there is only 1 submission kind, just run the checks
1. where there are multiple submission kinds allow the user to select the kind
1. if there is more than 1 kind AND at last one open collection allow the user to choose the kind in the context of the collection somehow, maybe just by listing `Kind (collectionA, collectionB)`?
1. a user should be able to check against a kind even if not available in an open collection

add command line parameters:
1. `curvenote check venue --kind <kind>`
1. `curvenote check venue --collection <collection>`
   - if more than on kind associated with that collection, prompt the user
   
Open to suggestions on the above proposal too.

stevejpurves:
note a good test case is to point to `staging` and `scipy` which currently has multiple kinds and collections.

fwkoch:
This makes sense, feels like it should basically just follow the same interface as `submit` (just more permissive on closed collections, etc).

stevejpurves:
fixed by #546

PR #546, opened 2024-04-17
✅ Allow collection/kind options on check

fwkoch:
This attempts to address https://github.com/curvenote/curvenote/issues/539

It reuses as much of the collection/kind selection logic from `submit` as possible. The biggest difference there is that it adds an option `allowClosedCollections` for the selection logic; for `submit` this is `false` (well, `undefined`) - keeping previous behaviour - but for `check`, it is `true`, allowing checks against closed collections. There's also some subtle change to language - a little more careful to not say "you are submitting to..." when just running checks.

By reusing the submit logic, it also means we (1) determine collection first then (2) determine kind; this differs slightly from the previous check behaviour (and the description in #539) where only `kind` is necessary. I think that's ok? Even if the `kind` is shared across collections, this is just an implementation detail. I think the user should always be thinking "I'm submitting to X collection's kind Y"

### DIFF
--- a/.changeset/tender-rules-hammer.md
+++ b/.changeset/tender-rules-hammer.md
@@ -1,0 +1,6 @@
+---
+"@curvenote/cli": patch
+"curvenote": patch
+---
+
+Allow collection/kind options on check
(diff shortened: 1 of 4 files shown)


stevejpurves:
tested this out locally with multiple collections & kinds, works great.
`submit` also behaving as expected!

PR #554, opened 2024-04-19
Consume updates to Curvenote's journals API

fwkoch:
Primarily, the Curvenote journals API has added a `/submissions/<id>/versions` endpoint on submissions; POST requests now go here instead of `/submissions/<id>`.


packages/curvenote-cli/src/submissions/utils.ts
```diff
@@ -82,7 +80,7 @@ export async function postNewWork(
   cdnKey: string,
   cdn: string,
   key: string,
-): Promise<LogInfo> {
+): Promise<WorkDTO> {
```
fwkoch, line 83: These functions which create works / work versions / submissions / submission versions now simply return the newly created object, rather than just the `LogInfo` (it's easier to build the log info in the containing functions where we pull the different parts together).

packages/curvenote-cli/src/submissions/submit.utils.ts
```diff
@@ -622,19 +624,19 @@ export async function createNewSubmission(
 
   submitLog.work = {
     id: work.id,
-    date_created: workResp?.date_created ?? work.dateCreated,
```
stevejpurves, line 625: The case changes not so much here but throughout, is that going to change the case of any of the fields in the SubmitLog and so potentially break build pages or other things dependent on the job payload/result blobs?

packages/curvenote-cli/src/submissions/utils.ts
```diff
@@ -92,53 +90,37 @@ export async function postNewWork(
...
     session.log.debug(`CDN key: ${cdnKey}`);
     session.log.debug(`Work Id: ${id}`);
     session.log.debug(`Work Version Id: ${version_id}`);
-    return {
-      id: id,
-      versionId: version_id,
-      dateCreated: date_created,
-      versionDateCreated: date_created,
-    };
+    return json;
```
stevejpurves, line 123: I like the simpler return and lack of remapping

packages/curvenote-cli/src/submissions/submit.utils.ts
stevejpurves: seems like it is independent of this: https://github.com/curvenote/curvenote/blob/f0eb01b1c8a5ad11a4fbcf00e46c9dda0a0c7b64/packages/curvenote-cli/src/submissions/submit.ts#L273 and https://github.com/curvenote/curvenote/blob/f0eb01b1c8a5ad11a4fbcf00e46c9dda0a0c7b64/packages/curvenote-cli/src/submissions/submit.utils.ts#L627

fwkoch: There should be no changes to the `submitLog` and therefore what is going to the `job` - the change is on the interface of the data we are _feeding into_ the log - i.e. we pull id and versions from api responses, rather from an intermediate remapped structure.

stevejpurves:
rebased on latest ahead of bringing this in

PR #557, opened 2024-04-22
directive tweaks

stevejpurves:
This PR tweaks/improves the `cn:articles` directive options

### DIFF
--- a/packages/curvenote-cli/.eslintrc.cjs
+++ b/packages/curvenote-cli/.eslintrc.cjs
@@ -1,4 +1,12 @@
 module.exports = {
   root: true,
   extends: ['curvenote'],
+  rules: {
+    'import/no-extraneous-dependencies': [
+      'error',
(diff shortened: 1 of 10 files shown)


packages/curvenote-cli/.eslintrc.cjs
```diff
@@ -1,4 +1,12 @@
 module.exports = {
   root: true,
   extends: ['curvenote'],
+  rules: {
+    'import/no-extraneous-dependencies': [
+      'error',
+      {
+        devDependencies: ['**/*.spec.ts'],
+      },
+    ],
```
rowanc1, line 10: This requires `vitest` to be in the dev deps of every package, which we don't want.

```
/* eslint import/no-extraneous-dependencies: ['off'] */
```

At the top of the spec file might be easier?

This doesn't trigger a lint warning anyways, just visually...

PR #559, opened 2024-04-22
🛠 fix submit issue due to stale dto access

stevejpurves:
* removed superfluous check
* getting sub.version.id from active_version_id
* updated types form @curvenote/common


stevejpurves:
don't need a changeset here as we're already releasing cli with the right info in the changelog

PR #562, opened 2024-04-22
🪓 split out `@curvenote/cli-plugin`

stevejpurves:
This is to allow lightweight import of the plugin & directives in to the `docs` site, for auto-documentation with the `myst-directive` plugin over there. 

This was not possible before as bringing in plugin from `@curvenote/curvenote-cli` caused various import ESM/dynamic import issues


packages/cli-plugin/.gitignore
```diff
@@ -0,0 +1,11 @@
+dist/
+temp/
+node_modules/
+.ipynb_checkpoints/
+_build/
+yalc.lock
+.yalc/
```
rowanc1: Probably fine at the top level? And this can be removed?

packages/cli-plugin/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "devDependencies": {
+    "concurrently": "^7.2.2",
+    "esbuild": "^0.15.12",
+    "eslint": "^8.21.0",
+    "eslint-config-curvenote": "latest",
+    "npm-run-all": "^4.1.5",
+    "prettier": "latest",
+    "rimraf": "^3.0.2",
+    "typescript": "^4.6.3",
+    "vitest": "^1.5.0"
```
rowanc1: All of these should just be in the workspace!

packages/cli-plugin/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "url": "git+https://github.com/curvenote/curvenote.git"
+  },
+  "bugs": {
+    "url": "https://github.com/curvenote/curvenote/issues"
+  },
+  "homepage": "https://github.com/curvenote/curvenote",
+  "scripts": {
+    "clean": "rimraf dist",
+    "dev": "tsc --project ./tsconfig.json --outDir dist --declaration -w --preserveWatchOutput --outDir dist -p .",
+    "lint": "eslint \"src/**/*.ts\" -c .eslintrc.cjs",
```
rowanc1: Can add `\"src/**/!(*.spec).ts\"` instead of all src files.

packages/cli-plugin/.eslintrc.cjs
```diff
@@ -0,0 +1,12 @@
...
+  root: true,
+  extends: ['curvenote'],
+  rules: {
+    'import/no-extraneous-dependencies': [
+      'error',
+      {
+        devDependencies: ['**/*.spec.ts'],
+      },
+    ],
+  },
```
rowanc1: This only works if vitest is in the devDeps, and I think that version mismatches are worse. My advice is to put:

```
eslint \"src/**/!(*.spec).ts\" -c ./.eslintrc.cjs
```

In the lint function, and ignore them from tests/CI.

Then if the red line still is annoying:

```
/* eslint-disable import/no-extraneous-dependencies */
```

rowanc1 (commented):
Comments on the lint and package split out, minor, but should be a little easier to handle the eslint versions etc.

Some of these are over in @fwkoch's PR, which actually re-enables the tests...

PR #563, opened 2024-04-23
👩‍💻 Update build cli to match myst, deprecate export cli

fwkoch:
- `curvenote build` now has the same interface as `myst build`
- `curvenote export` no longer exists - the underlying functions remain for legacy usge
- `curvenote auth list` was moved under `curvenote token` (and aliased) and some of the token functionality was improved (no more deleting all your tokens at once by accident!!)


rowanc1:
Tested locally, all seems to be good!

PR #565, opened 2024-04-24
🙋‍♂️ Just try to get venue collections for `check` without checking access

fwkoch:
Checks cannot be run against a journal if you are unauthenticated. For public journals, this does not need to be the case - even if you cannot `submit`, you should still be able to run the checks.

This PR separates `submit access check` from `getting the collections` - for `submit`, we still do both, but for `check` we only do the latter. For private sites, getting the collections should still fail, but for public sites, users can carry on with their checks.

### DIFF
--- a/.changeset/bright-jobs-matter.md
+++ b/.changeset/bright-jobs-matter.md
@@ -1,0 +1,5 @@
+---
+"@curvenote/cli": patch
+---
+
+Just try to get venue collections without checking access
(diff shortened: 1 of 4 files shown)


stevejpurves:
looks good! taking it for a spin locally

stevejpurves:
@fwkoch this looks good on both the `anon` and team access front. Found [this issue](https://github.com/curvenote/journals/issues/279) on the way but this is a site misconfiguration issue, rather than a client auth issue

PR #569, opened 2024-04-25
📓 Update check command to only need kind, not collection

fwkoch:
`curvenote check` will now run if only `--kind` is provided, or, if nothing is provided, it will only prompt for `kind`.

You may still specify `--collection`: If `collection` is specified and not `kind`, it will only prompt for kinds in the collection (this case is the previous behaviour, which used to require collection and kind). If both `collection` and `kind` are specified, only `kind` will be used for the checks, but there may be some additional error logging if collection and kind are incompatible.

### DIFF
--- a/packages/curvenote-cli/src/submissions/check.ts
+++ b/packages/curvenote-cli/src/submissions/check.ts
@@ -1,3 +1,4 @@
+import chalk from 'chalk';
 import { submissionRuleChecks } from '@curvenote/check-implementations';
 import { build } from 'myst-cli';
 import { logCheckReport, runChecks } from '../check/runner.js';
@@ -8,12 +9,14 @@ import type { Check } from '@curvenote/check-definitions';
 import {
   checkVenueExists,
   determineCollectionAndKind,
+  getSubmissionKind,
   getVenueCollections,
 } from './submit.utils.js';
+import { determineKind } from './kind.utils.js';
 
-//
-// get checks
-//
+/**
+ * Return list of checks from `kind` object and print log message
+ */
 export function prepareChecksForSubmission(
   session: ISession,
   venue: string,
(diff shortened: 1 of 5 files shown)


packages/curvenote-cli/src/submissions/check.ts
```diff
@@ -36,18 +39,72 @@ type CheckOpts = {
...
+          }
+          if (!collection.kinds.find((k) => k.id === kind.id)) {
+            session.log.info(
+              `${chalk.red(`Collection "${opts.collection}" does not support kind "${opts.kind}"`)}`,
+            );
+          }
+        }
+      }
+    } else if (opts.collection) {
+      const result = await determineCollectionAndKind(session, opts.venue, collections, {
```
fwkoch, line 98: This is the previous behaviour.

packages/curvenote-cli/src/submissions/check.ts
```diff
@@ -36,18 +39,72 @@ type CheckOpts = {
...
+        ...opts,
+        allowClosedCollection: true,
+      });
+      kind = result.kind;
+    } else {
+      kind = await determineKind(session, opts.venue, collections, opts);
```
fwkoch: This is a new function that determines only kind, based on venue.

packages/curvenote-cli/src/submissions/check.ts
```diff
@@ -36,18 +39,72 @@ type CheckOpts = {
...
-    const collections = await getVenueCollections(session, opts.venue);
-    const { kind } = await determineCollectionAndKind(session, opts.venue, collections, {
-      ...opts,
-      allowClosedCollection: true,
-    });
+    const collections = await getVenueCollections(session, opts.venue, false);
+    let kind: SubmissionKindDTO;
+    if (opts.kind) {
+      try {
+        kind = await getSubmissionKind(session, opts.venue, opts.kind);
```
fwkoch, line 73: Simplest case where `kind` is provided!

packages/curvenote-cli/src/submissions/submit.utils.ts
```diff
@@ -383,36 +381,55 @@ export async function checkVenueSubmitAccess(session: ISession, venue: string) {
...
+ * This will fail with `process.exit(1)` if the fetch for venue collections fails.
+ * By default, it also fails if there are no open collections.
+ *
+ * If `requireOpenCollections` is false, this function will not fail if there are
+ * only closed collections or no collections at all.
+ */
+export async function getVenueCollections(
+  session: ISession,
+  venue: string,
+  requireOpenCollections = true,
```
fwkoch, line 442: This parameter is new. This allows us to get a venue's collections, even if they are all closed or it's simply an empty list. Previously, both of those would cause `process.exit(1)`; now, that is still the default case, but setting `requireOpenCollections: false` allows both those cases to pass and still return the collections.

stevejpurves: ok cool and it is set to false in `check` but left as default `true` in `submit` 👍🏼

So in terms of the action, this means for no open collections the check will pass but then a submit will fail if collections are closed. seems reasonable / better!

stevejpurves:
❤️ 
image

stevejpurves:
one extension to this that would be awesome would be for the cli to give a hint at the shortcut after it runs...

e.g. `curvenote check scipy` followed by a kind selection results in the following being console logged after the checks.

> Hint: run this check directly using `curvenote submit scipy --kind article`

But if we want to get this out quick, we can leave that for later

stevejpurves:
@fwkoch tested it out locally and rebased

stevejpurves (approved):

fwkoch:
Added the hint you suggested @stevejpurves 
image

PR #595, opened 2024-06-11
Consume MyST toc changes

fwkoch:
This PR brings curvenote up to date with the latest myst commits - it adds support for project-config toc as well as the newly async config loading function.


packages/curvenote-cli/src/session/auth.ts
```diff
@@ -119,7 +119,7 @@ export async function setToken(log: Logger, token?: string) {
     ]);
     token = resp.token as string;
   }
-  const session = new Session(token, { skipProjectLoading: true });
+  const session = new Session(token);
```
stevejpurves, line 122: will that have an impact on `curvenote auth check` `curvenote token select`? i think the flag was there to stop project errors and warnings crop up when a user was interested in managing tokens. maybe this is handled in a better way elsewhere? or maybe even by design?

fwkoch: Yeah - I think this still behaves the same way. The reason for this change is a relatively tedious chain of events:
- Adding support for remote `extend` fields in the project config moved all functions that load configs from sync to async
- Previously, a `reload` happened on `Session` instantiation: https://github.com/curvenote/curvenote/blob/main/packages/curvenote-cli/src/session/session.ts#L207-L209
- Now, `reload` is async so it cannot happen in a constructor (at least as far as I can tell...)
- That means we need to call `new Session(...)` then `session.reload()` separately - I don't love this; rather than `reload` being the default behavior with an option to turn off, we now must remember to call it explicitly or potentially be stuck with some strange bugs.
- However, that means we no longer need the `skipProjectLoading` argument, and instead, everywhere that was `false` before we must call `reload` explicitly.
- Here, where we do want to skip loading, we simply don't call `reload` and it works as before...

stevejpurves: ok - thanks for the explanation, i was thinking something like that but wanted to be sure - and it's good to be aware of the new requirement on explicit `.reload()` 👍🏼

PR #650, opened 2024-10-24
🛣️ add bypass option and 🧹

stevejpurves:
this better enables local dev for themes


fwkoch (approved):
Pretty simple - just applies the same bypass logic from `config.json` to `myst.xref.json` and `myst.search.json`.

PR #665, opened 2025-01-22
🎟️ Upgraded token management and CLI configuration

stevejpurves:
- **🛻 token management and session overhaul**
- **✨ui tweaks**
- **🎟️ token overhaul complete**
- **☎️ cli tweaks**
- **👮‍♀️ no cdn changes needed - reverting**
- **👮🏾‍♂️ reverting last cli ui tweak**
- **📗 changeset**
- **🙈 skipping model tests for unauthenticated access**


packages/curvenote-cli/src/models.ts
```diff
@@ -121,7 +125,9 @@ export class User extends BaseTransfer<string, UserDTO> {
 
   $fromDTO = userFromDTO;
 
-  $createUrl = () => `/users/${this.id}`;
+  $createUrl = () => {
+    return `${this.session.config.editorApiUrl}/users/${this.id}`;
```
stevejpurves, line 129: Other existing models are all editor centric so we enforce that in their urls. Note `User` is not really used and `Team` will end up deprecated.

packages/curvenote-cli/src/models.ts
```diff
@@ -109,7 +109,11 @@ export class MyUser extends BaseTransfer<string, MyUserDTO> {
 
   $fromDTO = myUserFromDTO;
 
-  $createUrl = () => `/my/user`;
+  $createUrl = () => {
+    let audience = this.session.activeTokens.session?.decoded?.aud;
+    if (audience && !audience?.endsWith('/')) audience = audience.replace(/\/$/, '');
+    return `${audience ?? this.session.config.editorApiUrl}/my/user`;
```
stevejpurves, line 115: `MyUser` model is driven from the `audience`

packages/curvenote-cli/src/session/auth.ts
```diff
@@ -107,7 +111,7 @@ function updateCurrentTokenConfig(log: Logger, token?: string) {
  * If config file does exist with saved tokens, this token will be added
  * to the saved tokens and set as the current token.
  */
-export async function setToken(log: Logger, token?: string) {
+export async function setUserToken(log: Logger, token?: string) {
```
stevejpurves, line 42: being more explicit about User vs Session throughout

packages/curvenote-cli/src/session/auth.ts
```diff
@@ -119,25 +123,51 @@ export async function setToken(log: Logger, token?: string) {
     ]);
     token = resp.token as string;
   }
-  const session = new Session(token);
+
+  log.debug('Creating session with new token');
+  const session = await Session.create(token);
```
stevejpurves, line 56: `Session`s constructor is private, `await Session.create()` is the only way to create one.

packages/curvenote-cli/src/session/auth.ts
```diff
@@ -119,25 +123,51 @@ export async function setToken(log: Logger, token?: string) {
...
   if (!me.data.email_verified) throw new Error('Your account is not activated');
   const data = getTokens();
   const tokens = data.saved ? [...data.saved] : [];
   if (!tokens.find(({ token: t }) => t === token)) {
-    tokens.push({ api: session.API_URL, email: me.data.email, username: me.data.username, token });
+    tokens.push({
+      api: session.activeTokens.user?.decoded.aud ?? 'unknown-audience', // TODO this should be based on the audience
+      email: me.data.email,
+      username: session.activeTokens.user?.decoded.name ?? me.data.username ?? me.data.display_name,
+      note: session.activeTokens.user?.decoded.note,
```
stevejpurves, line 79: Exposing the string set by the user in the Admin UI

packages/curvenote-cli/src/session/utils.ts
```diff
@@ -30,10 +67,146 @@ export function getSession(opts?: SessionOpts & { hideNoTokenWarning?: boolean }
...
+
+/**
+ * makeDefaultConfig cerate the fallback configuration which only needs to be valid for
+ * the case where the user token was created on the legacy API
+ *
+ * @param session
+ * @param opts
+ * @returns
+ */
+export function makeDefaultConfig(audience: string): CLIConfigData {
```
stevejpurves: this is now the fallback when a token does not provide the `cfg` claim.

packages/curvenote/src/devhelp.ts
```diff
@@ -0,0 +1,27 @@
...
+import { clirun } from './clirun.js';
+import chalk from 'chalk';
+import type { ISession } from '@curvenote/cli';
+
+function devhelp(session: ISession) {
+  const message = `🚀 Curvenote CLI Developer Help 🚀
+
+    ${chalk.bold('Testing without uploading')}
+    
+    Run: DEV_CDN=true cn deploy -d
```
stevejpurves, line 11: because i always forget

packages/curvenote-cli/src/session/types.ts
```diff
@@ -9,7 +9,22 @@ export type SessionOpts = {
...
+  exp?: number;
+  iat?: number;
+  cfg?: string;
+  ignoreExpiration?: boolean;
+  name?: string;
+  note?: string;
+  [key: string]: any;
+};
+
+export type Token = { token: string; decoded: TokenPayload };
```
stevejpurves, line 38: we now store the payload and token in the session

packages/curvenote-cli/src/session/session.ts
```diff
@@ -216,16 +128,144 @@ export class Session implements ISession {
...
 
-  proxyAgent?: HttpsProxyAgent<string>;
-  _shownUpgrade = false;
-  _latestVersion?: string;
-  _jupyterSessionManagerPromise?: Promise<SessionManager | undefined>;
+  setUserToken(token: Token) {
+    this.$activeTokens.user = token;
+  }
+
+  async refreshSessionToken() {
```
stevejpurves, line 137: now explicit here

stevejpurves: but also every run gets a new session token, we could/should be caching them (future PR)

packages/curvenote-cli/src/session/session.ts
```diff
@@ -216,16 +128,144 @@ export class Session implements ISession {
...
+  }
+
+  async configure() {
+    if (!this.$activeTokens.session?.decoded.aud) return;
+    const { aud, cfg } = this.$activeTokens.session.decoded;
+    const audience = Array.isArray(aud)
+      ? aud[0]
+      : (this.$activeTokens.session.decoded.aud as string);
+
+    if (cfg) {
```
stevejpurves, line 225: if there is a `cfg` claim, hit that endpoint expecting config info

packages/curvenote-cli/src/session/session.ts
```diff
@@ -216,16 +128,144 @@ export class Session implements ISession {
...
+            ...headers,
+          },
+        });
+        if (response.ok) {
+          this.log.debug(`Response ok.`);
+          this.$config = (await response.json()) as CLIConfigData;
+          this.log.debug(`Configuration set: "${JSON.stringify(this.$config, null, 2)}".`);
+
+          // We are still setting this as some of the myst-cli functions rely on it
+          this.API_URL = this.$config?.editorApiUrl ?? 'INVALID';
```
stevejpurves, line 243: we need to keep setting this in order to keep some of the `myst-cli` things we are using happy (templatesUrl), works for now.

PR #674, opened 2025-01-27
🚩 improve the fallback venue error message env based token checks

stevejpurves:
Will print additional info for the "Venue not accepting" error message, and only uses thst in one place now. Additional improvements to ensure validaity and print additional information when CURVENOTE_TOKEN is used to set the token, as on CI.

### DIFF
--- a/packages/curvenote-cli/src/session/session.ts
+++ b/packages/curvenote-cli/src/session/session.ts
@@ -514,6 +514,10 @@ export async function getSession(
   let session;
   try {
     session = await Session.create(data.current, { logger });
+    if (data.environment) {
+      logger.warn('Checking user token...');
+      await checkUserTokenStatus(session);
+    }
   } catch (error) {
     logger.error((error as Error).message);
     process.exit(1);
--- a/packages/curvenote-cli/src/session/tokens.ts
+++ b/packages/curvenote-cli/src/session/tokens.ts
@@ -118,5 +118,15 @@ export function summarizeAsString({ note, username, email, api }: Omit<TokenData
 export function getCurrentTokenRecord(tokens?: ReturnType<typeof getTokens>) {
   const data = tokens ?? getTokens();
   if (!data.current) return;
+  if (data.environment) {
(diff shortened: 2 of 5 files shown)


packages/curvenote-cli/src/session/session.ts
```diff
@@ -514,6 +514,10 @@ export async function getSession(
   let session;
   try {
     session = await Session.create(data.current, { logger });
+    if (data.environment) {
+      logger.warn('Checking user token...');
+      await checkUserTokenStatus(session);
+    }
```
stevejpurves, line 520: This will now show up on CI and let us understand any token issues in the actions

packages/curvenote-cli/src/session/tokens.ts
```diff
@@ -118,5 +118,15 @@ export function summarizeAsString({ note, username, email, api }: Omit<TokenData
...
   if (!data.current) return;
+  if (data.environment) {
+    const { decoded } = decodeTokenAndCheckExpiry(data.current, chalkLogger(LogLevel.info));
+    return {
+      token: data.current,
+      api: decoded.aud,
+      email: decoded.email,
+      username: decoded.name ?? decoded.user_id,
+      note: 'From environment variable',
+    };
```
stevejpurves, line 129: tokens coming in via the env will look like any other token object

packages/curvenote-cli/src/submissions/submit.utils.ts
```diff
@@ -454,7 +449,7 @@ export async function getVenueCollections(
   const openCollections = collections.items.filter((c) => c.open);
 
   if (openCollections.length === 0) {
-    session.log.info(`${chalk.red(`🚦 venue "${venue}" is not accepting submissions.`)}`);
+    session.log.info(`${chalk.red(`🚦 venue "${venue}" has no open collections.`)}`);
```
stevejpurves, line 452: made this more specific, the "not accepting submissions." not only happens in 1 place and as a result of the `v1/sites/{name}/access` call.

PR #682, opened 2025-02-11
🔧 Update CDN functions from myst-theme and depend on these

fwkoch:
This copies in the latest versions of the util functions from `@myst-theme/common` and updates imports to pull from here rather than `myst-theme`.

For now, this is a bit aspirational - now, the functions here and in myst-theme are identical, but this allows us to upgrade curvenote separately first, then pull in changes to myst-theme later.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22161,17 +22161,6 @@
         "node": ">=16"
       }
     },
-    "packages/cdn/node_modules/@curvenote/common": {
-      "version": "0.2.13",
-      "resolved": "https://registry.npmjs.org/@curvenote/common/-/common-0.2.13.tgz",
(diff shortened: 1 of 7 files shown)


package-lock.json
```diff
@@ -22128,6 +22161,17 @@
...
+      "version": "0.2.13",
+      "resolved": "https://registry.npmjs.org/@curvenote/common/-/common-0.2.13.tgz",
+      "integrity": "sha512-6l5IMBMjdFvE8/2THdBdx3u4IxhIszYcRnq6mKT/yh7PxWzfpZzhSuynbnEKPP97JishMM7SpD71ELZMZ9tlfg==",
+      "license": "MIT",
+      "dependencies": {
+        "@curvenote/blocks": "^1.5.29",
+        "myst-common": "^1.7.4",
+        "myst-config": "^1.7.4"
+      }
+    },
```
rowanc1: This is expected?

rowanc1: i.e. it is nested?

fwkoch: This nesting is now gone (~though it's slightly buried since the `package-lock.json` diff is larger...~ no longer a big package-lock diff).

PR #693, opened 2025-03-18
🌍 Use `deploymentCdnUrl` for deployments.

stevejpurves:


packages/curvenote-cli/src/session/session.ts
```diff
@@ -246,7 +247,11 @@ export class Session implements ISession {
         });
         if (response.ok) {
           this.log.debug(`Response ok.`);
-          this.$config = (await response.json()) as CLIConfigData;
+          const configFromApi = (await response.json()) as CLIConfigData;
+          this.$config = {
+            ...defaultConfig, // ensure the deploymentCdnUrl is set
+            ...configFromApi,
```
stevejpurves, line 253: Allows us to lean on local config for `deploymentCdnUrl` until the API catches up

packages/curvenote-cli/src/web/deploy.ts
```diff
@@ -51,7 +51,7 @@ export async function deploy(
 
   let cdnKey: string;
   if (!process.env.DEV_CDN || process.env.DEV_CDN === 'false') {
-    const result = await uploadToCdn(session, session.config.publicCdnUrl, opts);
+    const result = await uploadToCdn(session, session.config.deploymentCdnUrl, opts);
```
stevejpurves, line 54: no longer deploying to `pub`

PR #694, opened 2025-03-20
📆 Set date published and 🔧 Use both /my/ and /sites/ APIs for finding submissions

fwkoch:
This PR includes 2 totally separate features 😕:

1. `date_published` may now be set explicitly the first time a submission is published from the CLI. If it is not set, the frontmatter `date` will be used. If that doesn't exist, it will fall back to the original behaviour of using "today"
2. the Curvenote API has changed so users may no longer access their submissions to a site via `/sites/<site>/submissions` - that endpoint is restricted to site admins. This means when checking for existing submissions to a site we also must use `/my/submissions`.

### DIFF
--- a/packages/curvenote-cli/src/submissions/submit.ts
+++ b/packages/curvenote-cli/src/submissions/submit.ts
@@ -102,7 +102,7 @@ export async function submit(session: ISession, venue: string, opts?: SubmitOpts
       const exists = await checkForSubmissionKeyInUse(session, venue, key);
       if (exists) {
         session.log.warn(
-          `⛔️ This work has already been submitted to "${venue}", but you don't have permission to access that submission.`,
+          `⛔️ This work has already been submitted to a Curvenote site, but you don't have permission to access that submission.`,
         );
         session.log.info(
           'If you still want to make a new submission, you may explicitly add flag "--new"',
--- a/packages/curvenote-cli/src/submissions/submit.utils.ts
+++ b/packages/curvenote-cli/src/submissions/submit.utils.ts
@@ -525,10 +525,21 @@ export async function getAllSubmissionsUsingKey(
       session,
       `/my/submissions?key=${key}`,
     );
+    // These extra fetches are required for the old version of the API where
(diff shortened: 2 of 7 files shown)


packages/curvenote-cli/src/submissions/submit.ts
```diff
@@ -102,7 +102,7 @@ export async function submit(session: ISession, venue: string, opts?: SubmitOpts
       const exists = await checkForSubmissionKeyInUse(session, venue, key);
       if (exists) {
         session.log.warn(
-          `⛔️ This work has already been submitted to "${venue}", but you don't have permission to access that submission.`,
+          `⛔️ This work has already been submitted to a Curvenote site, but you don't have permission to access that submission.`,
```
fwkoch, line 105: This is just an improvement to user feedback - Turns out, this API endpoint checks all venues all the time. That means this warning arises if the work has been submitted anywhere.

packages/curvenote-cli/src/submissions/submit.utils.ts
```diff
@@ -510,14 +510,42 @@ export async function getAllSubmissionsUsingKey(
...
+      session,
+      `/my/submissions?key=${key}`,
+    );
+    // These extra fetches are required for the old version of the API where
+    // the 'key' query parameter is not respected on /my/submissions.
+    // This may be removed (along with the 'correctKey' check below) once
+    // the API is updated.
+    const works = await Promise.all(
+      mySubmissions.items.map((sub) => {
+        return getFromUrl(session, sub.links.work);
```
fwkoch, line 534: Since `key` is only on `work`, not `submission`, these extra API calls are required to filter the submission list.

On the next release, the API will be updated to correctly respect the `key` query on `/my/submissions` and we can delete these calls.

packages/curvenote-cli/src/submissions/submit.ts
stevejpurves: I remember, it was misleading.

stevejpurves (approved):

PR #705, opened 2025-05-05
🙈 Enable anonymous api GET requests to enable public venue `checks`

fwkoch:
When unauthenticated `curvenote check` runs some default hardcoded checks, but if a user wants to run venue-specific checks `curvenote check <venue>` fails, even if the venue is public. This is because all journal api requests are prevented when no token is set.

The CLI messaging made this particularly problematic: right at the beginning, there is a "you are unauthenticated" warning, then the build/checks seem to continue, then it says "venue not found" because there is a try/catch around the api request failure. At first, I was thinking we need to update that to "you must be authenticated to run checks against venue" so the user knows the problem is auth, not incorrect venue.

However, then I was questioning - Why can't we just run checks against public venues, even when unauthenticated? That would allow users to run `curvenote check` sooner.

So, in this PR, I'm enabling unauthenticated `checks` by allowing unauthenticated `GET`s to the journal api.

### DIFF
--- a/packages/curvenote-cli/src/session/session.ts
+++ b/packages/curvenote-cli/src/session/session.ts
@@ -109,9 +109,8 @@ export class Session implements ISession {
         session.setUserToken({ token, decoded });
         await session.refreshSessionToken();
       }
-
-      await session.configure();
     }
+    await session.configure();
 
     return session;
   }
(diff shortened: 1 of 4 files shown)


packages/curvenote-cli/src/session/session.ts
```diff
@@ -109,9 +109,8 @@ export class Session implements ISession {
         session.setUserToken({ token, decoded });
         await session.refreshSessionToken();
       }
-
-      await session.configure();
     }
+    await session.configure();
```
stevejpurves, line 113: Yeah, I think one issue here is now when unauthenticated, it's going to fall back to the default config, Which will always be prod probably OK for now, but it could cause issues for non-curvenote public deployment.

stevejpurves (approved):
We could go with these changes, they just feel a bit isolated, with the same logic, not applicable to build and start they could run over anonymous sessions if they don't already. 

Definitely we should be able to check against public venues.

fwkoch:
Yeah, we will need to improve things as we start white-labeling our client for other deployments, potentially. Bringing this in for now though!

PR #712, opened 2025-06-26
Update CLI for work-based site landing content

fwkoch:
- Consume myst `plugin` loading changes
- Remove `content` from `SiteDTO` (there is a new `SiteWithContentDTO` too)
- Add `curvenote work push` for directly uploading a new version of a work
- Add `curvenote site init <siteName> --set-content` to set the current work as site landing content

- [ ] TODO: on `push` also push `source/` folder alongside built site content.


package.json
```diff
@@ -30,7 +30,7 @@
     "prettier": "latest",
     "rimraf": "^3.0.2",
     "turbo": "latest",
-    "typescript": "latest",
+    "typescript": "5.6.2",
```
fwkoch, line 33: We are threading the needle here a bit: `<5.5` produces a type error in `myst-cli`; `>5.8` produces an error with `markdown-it-myst`. For now, I am pinning to the version used in MyST.

packages/curvenote-cli/src/sites/utils.ts
```diff
@@ -0,0 +1,55 @@
+import chalk from 'chalk';
+import inquirer from 'inquirer';
+import type { ISession } from '../session/types.js';
+import { getFromJournals } from '../utils/api.js';
+
+export async function ensureVenue(
```
fwkoch, line 6: I did quite a bit of moving functions around. Previously, a ton of helper functions were under `submissions/` even if they could be used in non-submission context. Now, these functions are moved under `sites/` and `works/` where appropriate.

packages/curvenote-cli/src/sites/utils.ts
```diff
@@ -0,0 +1,55 @@
...
+  session: ISession,
+  venue: string | undefined,
+  opts: { yes?: boolean; action?: string } = { action: 'submit' },
+) {
+  if (venue) return venue;
+  if (opts?.yes) {
+    throw new Error(`⛔️ Site must be specified to continue`);
```
fwkoch, line 13: Also, I started removing "Venue" from user-facing messages; previously, this message was "Venue must be specified..."

packages/curvenote-cli/src/submissions/submit.ts
```diff
@@ -72,27 +56,23 @@ export async function submit(session: ISession, venue: string, opts?: SubmitOpts
...
-  let key = workKeyFromConfig(session);
+  // Determine Work key
+  let inputKey = workKeyFromConfig(session);
   // Deprecation step to handle old transfer.yml files
-  key = (await keyFromTransferFile(session, venue, key, opts)) ?? key;
-
-  if (!key) {
-    key = await promptForNewKey(session, opts);
-    await writeKeyToConfig(session, key);
+  inputKey = (await keyFromTransferFile(session, venue, inputKey, opts)) ?? inputKey;
```
stevejpurves, line 62: i guess this is just getKeyFromConfig these days as transfer files were deprecated ages ago? if not that's a code branch to squeeze out later.

stevejpurves (approved):
It's good so far. @fwkoch I think you said you'd still be building on top of this? For the source files upload?

packages/curvenote-cli/src/submissions/submit.ts
fwkoch: Yes - if we are ready to delete all the legacy transfer file stuff, we could then create a nice reusable function with:

```typescript
// Determine Work key
let inputKey = workKeyFromConfig(session);
if (!inputKey) {
  inputKey = await promptForNewKey(session, opts);
  await writeKeyToConfig(session, inputKey);
}
exitOnInvalidKeyOption(session, inputKey);
```

With the transfer file stuff, it's a little more annoying, since that comes in for `submissions` but not for `works`.

Anyway - probably time to clean up transfer files, but separate PR.

fwkoch:
Thanks @stevejpurves - since you've reviewed this, I will merge it and create a separate, smaller PR with the `source/` folder stuff.

PR #714, opened 2025-06-27
🎁 Write source folder to cdn when pushing work

fwkoch:
This reuses the MECA build process to create a folder with `source` files (pulled out of the MECA `bundle/` folder). This new `source/` folder gets added to the `site/` folder and uploaded during `curvenote work push`.

### DIFF
--- a/packages/curvenote-cli/src/works/utils.ts
+++ b/packages/curvenote-cli/src/works/utils.ts
@@ -2,6 +2,7 @@ import { v4 as uuid } from 'uuid';
 import inquirer from 'inquirer';
 import fs from 'fs/promises';
 import type { ExportWithOutput } from 'myst-cli';
+import { silentLogger } from 'myst-cli-utils';
 import { ExportFormats } from 'myst-frontmatter';
 import AdmZip from 'adm-zip';
 import {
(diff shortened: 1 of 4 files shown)


packages/curvenote-cli/src/works/utils.ts
```diff
@@ -23,6 +28,76 @@ export const CDN_KEY_RE =
...
+    if (!projectFile) {
+      activeLogger.debug('No project file found');
+      return;
+    }
+    const mecaExport: ExportWithOutput = {
+      format: ExportFormats.meca,
+      output: join(session.buildPath(), 'temp', 'meca-export.zip'),
+      articles: [],
+    };
+    await runMecaExport(session, projectFile, mecaExport, { projectPath });
```
stevejpurves, line 56: so this will run even though there may not be a `meca` key/setup in the project frontmatter?

packages/curvenote-cli/src/works/utils.ts
```diff
@@ -23,6 +28,76 @@ export const CDN_KEY_RE =
...
+      await fs.mkdir(dirname(destPath), { recursive: true });
+      if (!entry.isDirectory) {
+        await fs.writeFile(destPath, entry.getData());
+      }
+    }
+    activeLogger.info(`🎁 Source folder created`);
+    (session as any).$logger = activeLogger;
+
+    // Clean up the temporary MECA zip
+    await fs.unlink(mecaZipPath);
```
stevejpurves, line 94: 👍🏼

stevejpurves (approved):
Looks great. I was just double checking to make sure that it didn't rely on the article (that you are testing with) being setup with meca export in it's frontmatter.

packages/curvenote-cli/src/works/utils.ts
fwkoch: Yes, that's correct. The input `mecaExport` here is handcrafted to dump a MECA file to the `temp` folder; it does not use any of the `collectExportOptions` infrastructure to pull in `exports` from project/page frontmatter.

That means even if there _is_ a `meca` export setup for the project, this will ignore that and create a new one. I think that's good for simpler code and more consistent outputs.

stevejpurves (approved):
looks good @fwkoch - so we'll expect the source mats to appear in `source/` on the `cdn`.

PR #728, opened 2025-08-13
🔌 Load built-in plugins by default

rowanc1:


packages/curvenote-cli/src/session/session.ts
```diff
@@ -369,8 +369,7 @@ export class Session implements ISession {
   _pluginPromise: Promise<ValidatedCurvenotePlugin> | undefined;
 
   async loadPlugins(plugins: PluginInfo[]) {
-    const loadedPlugins = await loadProjectPlugins(this, plugins);
-    this.plugins = combinePlugins([getBuiltInPlugins(), loadedPlugins]);
+    this.plugins = await loadProjectPlugins(this, plugins);
```
rowanc1, line 372: these are now additive as well, i.e. taking the existing plugins on the session.

PR #733, opened 2025-10-10
feat/change nested project handling

stevejpurves:
https://www.loom.com/share/f1902825456b48ae8a685b976cffd8b1?sid=20b22f92-311f-44de-81e4-7b0bd39a5e61


fwkoch (approved):
This makes sense - gives users an option to avoid the previous, implicit, multi-project site. But keep the defaults.

I see all this getting cleaned up once we implement the mythical "one-project, one-site, only-run-in-project-root" cleanup in myst...

PR #740, opened 2025-11-19
👊 myst v1.6.6

stevejpurves:


stevejpurves:
basics are tested and working and build time execute is working, which was related to the mystmd changes

PR #760, opened 2026-01-07
Add Husky for pre-commit checks

stevejpurves:
@fwkoch @rowanc1 brought husky in to catch unwanted .app-config.schema changes


types/app-config.d.ts
```diff
@@ -361,12 +361,159 @@ export interface VideoElement {
...
 
 export interface Extensions {
+  /**
+   * HHMI Compliance Extension Module Configuration
+   */
+  "hhmi-compliance"?: HHMIComplianceExtensionModule;
+  /**
+   * PMC Extension Module Configuration
+   */
+  pmc?: PMCExtensionModule;
```
stevejpurves: hmm but they can still creep in here

PR #775, opened 2026-01-21
🧑‍💻 Enable signup on SCMS and provision editor accounts for new users

fwkoch:
- adds `config.api.createEditorUser` flag to the config
- if true and if we are provisioning new users, this will check if firebase/editor accounts exist - if they do, it will redirect the user to login with those accounts, and if not, it will create those accounts when signup is complete.
- when signing up, if email comes from the auth provider, you cannot change this. We are still missing email verification for orcid signup on SCMS.

---

> [!NOTE]
> Implements Editor account provisioning and hardens auth flows.
> 
> - Adds `createEditorUser` option; `completeSignupFlow` now creates Editor Firestore users and, for non-Firebase signups, creates a Firebase Auth user, handling email collisions by deleting pending users and redirecting to login
> - New helpers: `createEditorUser` (Firestore), `generateRandomPassword`, `dbDeletePendingUser`; Firebase server now exposes `getServerFirestore`; centralized `createLogoutHeaders` and used in logout route and error redirects
> - Auth strategies: Google linking logic streamlined; ORCID prevents linking to a different user and redirects if a Firebase user already exists for the email; improved analytics/events
> - UI: preferred login shows all providers with Firebase password form; email field becomes read-only when sourced from auth; avatars wrapped in `ClientOnly`
> - Types/config: add `createEditorUser` to Firebase/ORCID client-side options; minor copy tweaks; dependency change for `@curvenote/scms-sites-ext`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7268700e906ff23f281ae35082ba66bfbbd50398. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


platform/scms/app/routes/app/settings.linked-accounts/route.tsx
```diff
@@ -93,6 +93,8 @@ export default function LinkedAccounts({ loaderData }: Route.ComponentProps) {
   );
   const showDivider = linkedProviders.length > 0 && linkableOrPendingProviders.length > 0;
 
+  // TODO: Handle error toasts from URL parameters (e.g., from account linking failures)
```
fwkoch, line 96: I thought this would be easy - I added an error redirect that adds the "toast" query params to this URL, but couldn't get the toast to work...

fwkoch:
I've tested this with:

- orcid signup / existing editor account -> kicked out to signup with google (the pending orcid user is deleted when the existing email is specified in the signup form)
- orcid signup / no editor account / no scms account linked to orcid -> scms + editor accounts created
- google signup / existing editor account -> this is just a sign-in
- google signup / no editor account -> scms + editor accounts created

packages/scms-server/src/backend/signup/index.server.ts
```diff
@@ -93,6 +102,77 @@ export async function completeSignupFlow(
...
+    let firebaseUid: string = user.id; // Default to user.id (for Firebase users, this is already the Firebase UID)
+
+    // For non-Firebase users, we need to create a Firebase Auth user
+    if (user.primaryProvider !== 'google' && user.primaryProvider !== 'firebase-email') {
+      const serverAuth = await getServerAuth();
+      // Check if Firebase Auth user already exists by email
+      let firebaseUser;
+      let firebaseUserExists = false;
+      try {
+        firebaseUser = await serverAuth.getUserByEmail(email);
```
stevejpurves, line 118: This branch means that the user has not logged in via google, and has not logged in with firebase-mail. So this email comes from either another provider OR was supplied during the signup flow and is unverified by us.

So this is fine as we are going to throw and insist the user logs in independently 👍🏼. The only way to do this differently would be to verify the email with our own verification flow, correct? (that's a future thing)

packages/scms-server/src/backend/signup/index.server.ts
```diff
@@ -93,6 +102,77 @@ export async function completeSignupFlow(
...
+        if (error?.code === 'auth/user-not-found') {
+          // Create new Firebase Auth user
+          const password = generateRandomPassword();
+          firebaseUser = await serverAuth.createUser({
+            email,
+            displayName: dataCollectionStep?.displayName,
+            emailVerified: true, // Assume verified if they completed signup
+            password, // Random password - user can reset via "forgot password"
+          });
+          firebaseUid = firebaseUser.uid;
```
stevejpurves, line 130: ok this gets our new firebase user identity

packages/scms-server/src/backend/signup/index.server.ts
```diff
@@ -93,6 +102,77 @@ export async function completeSignupFlow(
...
+        throw redirect(
+          '/login?error=true&message=An account with this email already exists. Please log in using your existing Editor account.',
+          { headers },
+        );
+      }
+    }
+
+    // Create Editor Firestore user
+    try {
+      await createEditorUser(firebaseUid, {
```
stevejpurves, line 152: and this uses the new identity to create a new user object

packages/scms-server/src/backend/signup/index.server.ts
```diff
@@ -93,6 +102,77 @@ export async function completeSignupFlow(
...
+        email,
+        displayName: dataCollectionStep?.displayName,
+        emailVerified: true,
```
stevejpurves, line 155: i guess we are just bypassing this for now. If we left this false they would be challenged when signing for the editor? @fwkoch did you try this with false by any chance? just curious

packages/scms-server/src/modules/auth/orcid/register.server.ts
```diff
@@ -94,6 +95,14 @@ export function registerOrcidStrategy(config: AppConfig, auth: Authenticator<Aut
         let dbUserViaOrcid = await dbGetUserByLinkedAccount('orcid', idPayload.sub);
 
         if (dbUserViaOrcid) {
+          // Check if this ORCID account is already linked to a different user
+          const user = session.get('user');
+          if (user && dbUserViaOrcid.id !== user.userId) {
+            // ORCID account is already linked to a different user
+            throw redirect(
+              `/app/settings/linked-accounts?error=true&provider=orcid&message=${encodeURIComponent('This ORCID account has already been linked to another account.')}`,
+            );
```
stevejpurves: interesting @fwkoch is ORCID the only account we do this for? conceivably this is possible for any account type that we are trying to link

stevejpurves: hmm in google at least it seems this is covered by a different flow

stevejpurves: actually, no I think we need to look at that this is specifically for the linking case where we find a different user based on the linked account, we still need to check that the account that we are authenicating is not linked to another user. could you take a look at that for the other providers @fwkoch

stevejpurves (commented):
a bunch of comments, but one of the cases that you have guarded for in the ORCID account linking I don't see why we don't also have to have that in place in the other providers when linking too

packages/scms-server/src/backend/signup/index.server.ts
fwkoch: Ah, yes. If we see they already have a firebase account, they must use that. They are kicked out of this flow.

I think I agree with what you are saying - right now, we are a little extra defensive because we don't have email verification for ORCID login; if someone logs in with ORCID, but ORCID does not share their email, they can type whatever email. So we don't want to allow hijacking existing accounts. Once we have the flow to verify email, we could just connect all accounts on the same email that we find in our DB.

packages/scms-server/src/modules/auth/orcid/register.server.ts
fwkoch: I will add this for other providers.

It feels slightly different for google (getting back to our verify flow) - If someone else uses your email for their ORCID login, that email is not verified. Then if you log in with google for real, you won't be able to create an account because someone already has an erroneous account with your email.

I still think it's better to do this check, rather than potentially muddying our DB with multiple accounts on the same email (or whatever else happens now...). But we should probably tighten up the email verification for ORCID soon.

packages/scms-server/src/backend/signup/index.server.ts
fwkoch: Hmm, I didn't try it with `false`. It might be totally fine to do that, and just get the challenge on the editor side?

I didn't really even think about it - my thinking was just: we've already done the signup flow on scms, can just setup editor user as complete 🤷

PR #844, opened 2026-04-06
🦸 Remove duplicate hero directive

rowanc1:


rowanc1:
I haven't updated the package-lock files.

PR #859, opened 2026-04-21
🧑‍🚀 dynamic system roles

stevejpurves:
image

---

> [!NOTE]
> **High Risk**
> Changes core authorization behavior and database schema (including removing a `SystemRole` enum value), so misconfiguration or migration issues could inadvertently grant/deny access across the app.
> 
> **Overview**
> Moves **system role → scope mappings** from hardcoded server logic into the database via a new `SystemRoleScope` table, and updates scope resolution (`getUserById`, `userHasScope`, admin guards, and “list all sites” behavior) to prefer `user.system_scopes` loaded from DB with a fallback to in-code defaults.
> 
> Adds a platform-admin UI under `/app/system/roles` to view/update/delete per-system-role scope lists (including extension scopes), logs these changes as `ROLE_UPDATED` activities, and seeds default mappings; also introduces new app-level *feature* scopes (`app:*:feature`) and switches key routes (dashboard/settings/sites) to gate on those.
> 
> Removes the legacy `SystemRole.PLATFORM_ADMIN` enum value (and related UI/test token plumbing), shifting platform admin access to scope/role assignment rather than a dedicated system role.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3575934b50d8d7f9fc1dcffa105afdc6eef90d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/scms-server/src/backend/roles.server.ts
```diff
@@ -1,29 +1,64 @@
...
+      scopes: true,
+      date_created: true,
+      date_modified: true,
+    },
+  });
+  if (!row) {
+    return {
+      role,
+      scopes: getDefaultSystemRoleScopes(role),
+      fallback_scopes: getDefaultSystemRoleScopes(role),
```
stevejpurves: fallback_scopes is a pure ui concern, system.roles should get that itself and we should not be returning it here from a function used on the critical auth path

PR #901, opened 2026-05-25
ETL endpoint (and leftover fixes to CLI register workflow)

fwkoch:
This exposes a single ETL endpoint for bulk work registration by site admin.

It also includes changes made to the `register` CLI workflow, not relevant to the new API endpoint, but still potentially useful for end-users of the CLI.

### DIFF
--- a/packages/curvenote-cli/src/sites/utils.ts
+++ b/packages/curvenote-cli/src/sites/utils.ts
@@ -39,17 +39,14 @@ export function venueQuestion(session: ISession, action = 'submit') {
 
 /**
  * Ensure that a `site` exists by performing a basic request to the Site
- *
- * If Site does not exist, fails with `process.exit(1)`.
  */
 export async function checkVenueExists(session: ISession, venue: string) {
   try {
     session.log.debug(`GET from journals API /sites/${venue}`);
     await getFromJournals(session, `/sites/${venue}`);
     session.log.debug(`found Site "${venue}"`);
-  } catch (err) {
-    session.log.debug(err);
+  } catch {
     session.log.error(`${chalk.red(`😟 Site "${venue}" not found.`)}`);
-    process.exit(1);
+    throw new Error(`Site "${venue}" not found.`);
   }
 }
--- a/packages/curvenote-cli/src/works/create.ts
+++ b/packages/curvenote-cli/src/works/create.ts
@@ -9,7 +9,12 @@ import fs from 'node:fs';
 import path from 'node:path';
 import { load as yamlLoad } from 'js-yaml';
 import { getFromJournals } from '../utils/api.js';
-import type { MySubmissionsListingDTO, WorkDTO } from '@curvenote/common';
+import type {
+  CollectionDTO,
+  MySubmissionsListingDTO,
+  SubmissionKindDTO,
+  WorkDTO,
+} from '@curvenote/common';
 import { resolveExistingWork } from './resolveExistingWork.js';
 
 function detectContentYamlPath() {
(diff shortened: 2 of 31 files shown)


packages/curvenote-cli/src/sites/utils.ts
```diff
@@ -39,17 +39,14 @@ export function venueQuestion(session: ISession, action = 'submit') {
...
   try {
     session.log.debug(`GET from journals API /sites/${venue}`);
     await getFromJournals(session, `/sites/${venue}`);
     session.log.debug(`found Site "${venue}"`);
-  } catch (err) {
-    session.log.debug(err);
+  } catch {
     session.log.error(`${chalk.red(`😟 Site "${venue}" not found.`)}`);
-    process.exit(1);
+    throw new Error(`Site "${venue}" not found.`);
```
fwkoch, line 50: This first batch of changes replaces `proces.exit(1)` with `throw` (which is then caught in the CLI wrapper).

This feels better for the `curvenote-cli` library which can be imported independent of the actual CLI. In that context, `process.exit(1)` will cause an abrupt shutdown that bypasses all error handling.

The side effect of this change though, as currently implemented, is, in the context of the CLI, we will now get duplicate errors printed: One printed here, another printed when the CLI wrapper catches the error. 

Now, with the new ETL endpoint, these `process.exit` changes are less relevant, so if we do not want to worry about it, we can just revert them.

packages/curvenote-cli/src/works/create.ts
```diff
@@ -21,6 +26,18 @@ function detectContentYamlPath() {
   return undefined;
 }
 
+function resolveContentYamlPath(opts?: RegisterWorkOpts): string | undefined {
```
fwkoch, line 29: This next batch of changes improves `work.register` from the CLI. These include allowing an explicit yaml config file path...

packages/curvenote-cli/src/works/create.ts
```diff
@@ -70,14 +87,61 @@ function parseMetadataJson(metadataInput?: string) {
...
+>();
+
+function registrationTargetsCacheKey(
+  venue: string,
+  opts?: Pick<RegisterWorkOpts, 'collection' | 'kind'>,
+): string {
+  return `${venue}\0${opts?.collection ?? ''}\0${opts?.kind ?? ''}`;
+}
+
+async function resolveRegistrationTargets(
```
fwkoch, line 123: ...early resolution of kind/collection...

packages/curvenote/src/works.ts
```diff
@@ -50,7 +54,13 @@ function makeWorksRegisterCLI(program: Command) {
     .addOption(new Option('--source <string>', 'Source label that is written to work.contains'))
     .addOption(
       new Option(
-        '--metadata <json-or-file>',
+        '--work-metadata <json-or-file>',
```
fwkoch, line 57: ...and explicit work version and submission version metadata.

All of the above three changes, while they are again not relevant with the ETL endpoint, I think are still worth keeping.

packages/scms-server/src/backend/etl/register-work.server.ts
```diff
@@ -0,0 +1,423 @@
...
+          status: 'PENDING',
+          work_version: { connect: { id: workVersionId } },
+          kind: kindConnect(siteId, kindName),
+        },
+      });
+    }
+  });
+}
+
+export async function etlRegisterWork(
```
fwkoch, line 336: This is the ETL handler. Now, a single register uses ~6 DB round-trips: 3 for site + auth, 1 for doi lookup, 1 for tag check, then the large "create" transaction.

packages/scms-server/src/backend/etl/register-work.server.ts
```diff
@@ -0,0 +1,423 @@
...
+  if (exact) return exact.id;
+
+  const norm = normalizeDoiForCompare(doi);
+  const candidates = await prisma.work.findMany({
+    where: {
+      doi: { not: null },
+      ...ownerFilter,
+    },
+    select: { id: true, doi: true },
+    take: 4,
```
stevejpurves, line 98: @fwkoch why take 4 here?

packages/scms-server/src/backend/etl/auth.server.ts
```diff
@@ -0,0 +1,91 @@
...
+        id: true,
+        name: true,
+        external: true,
+        site_users: {
+          where: { user_id: userId, role: SiteRole.ADMIN },
+          select: { id: true },
+          take: 1,
+        },
+      },
+    }),
```
stevejpurves, line 77: we could probably kill the user query and nest that in the site query which would resolve by a join instead BUT that is not going to change the amount of data returned and i'm not 100% sure the prisma will definitely compact it to 1 query

PR #913, opened 2026-05-27
📝 Increase work title length

fwkoch:
Just increases title length validation from 255 to 500. The example biorxiv articles I inspected that failed this validation had lengths of ~260-280, so hopefully 500 will be more than satisfactory.


fwkoch:
We should do a validation pass to normalize title (and other fields) across POST creation, UI updates, form metadata, uploads processing, etc. But I didn't want to increase the blast of this PR outside the tiny validation change that is causing failures.

PR #935, opened 2026-06-09
🧑‍🎓 Add submission endpoints outside site scope

fwkoch:
This PR adds 2 new API endpoints:

`/v1/submissions` - this lists all published submissions on all public sites, following the same logic and query params as the site-scoped `/v1/sites/<site>/works` endpoint. You can add `site` queries to limit to one or more public sites on the top-level `/v1/submissions` endpoint

`/v1/doi/<prefix>/<suffix>` - this returns a public submission based on doi, regardless of site. This also parallels a site-scoped endpoint, `/v1/sites/<site>/doi/<prefix>/<suffix>`, with the same queries (e.g. `tag`). It also includes a `site` query param in case one DOI exists on multiple sites, so you can specify which one.

Both of these endpoints will enable our theme to serve submissions from multiple sites.


docs/planning/global-catalog-listing-performance.md
```diff
@@ -0,0 +1,672 @@
+# Plan: Global Public Catalog Performance (`/v1/submissions`, `/v1/doi`)
+
```
stevejpurves, line 2: worth a read -- and Phase 1 should be doable and might work just as well as what we have on a single site now.

Phase 2 is for ultimate performance but is much higher complexity with a lot of business logic going into db triggers

stevejpurves:
@fwkoch I think we are going to back off on this approach for now, and stick to the per-site endpoints we have for the next reader phase. This PR has been good though to surfaces some of the performance challenges around a set of cross site endpoints.

PR #945, opened 2026-06-12
🔧 Handle re-extracted articles on etl endpoint

fwkoch:
Previously, when registering an article via our ETL endpoint it looked to see if `doi` and `versionTag` already existed in the DB. If so, it skipped registration. If an article was updated in place - no new `doi` or `versionTag`, just modified content under a new `cdn_key`, the registration would still skip.

Now, the ETL endpoint enables re-registering of a single `doi` and `versionTag` if the `cdn_key` does not match. In order for this to happen, a few criteria need to be met:
- the client calling the ETL API must specify `allow_retagging: true` - that means this behavior is opt-in
- the `cdn_key` cannot match _any_ versions of the work, not just the currently-tagged version - this prevents accidentally re-tagging to a correct new version then accidentally re-tagging back to the old version again
- the `cdn_key` must have a matching prefix as the currently registered work - this prevents a totally different article from accidentally "stealing" the `doi`/`versionTag`

When re-registering a submission version, the metadata is as follows. Given an old submission version:

```
{
  "id": "<old-submisson-version-id>"
  "tags": ["v1"],
  "metadata": {
    "biorxiv": {
      "article_id": "…",
      "source_key": "…",
      "version": "v1"
    }
  }
}
```
When a new submission version is re-registered with the same `doi`/`versionTag`, the two submissions will look like:
```
{
  "id": "<old-submisson-version-id>"
  "tags": [],
  "metadata": {
    "biorxiv": {
      "article_id": "…",
      "source_key": "…",
      "version": "v1",
      "superseded_by_submission_version_id": "<new-submission-version-id>",
      "superseded_at": "2026-06-10T12:00:00Z"
    }
  }
}

{
  "id": "<new-submisson-version-id>"
  "tags": ["v1"],
  "metadata": {
    "biorxiv": {
      "article_id": "…",
      "source_key": "…",
      "version": "v1",
      "supersedes_submission_version_id": "<old-submission-version-id>"
    }
  }
}
```

Note, the new version points backwards and the old version points forwards. They both keep the `v1` version in `metadata`, but only the new version keeps `v1` in the `tags`.


stevejpurves (commented):
looks good @fwkoch, some thoughts

* Should we create an Activity on this change? TAGGED_VERSION_CHANGE or similar
* is it worth using shorter property names in json fields where we can? (saving bytes versus high readability)

(Note I really want to get the activity type out of enums to make it easier to fire off various activites, but not in this PR)

fwkoch:
Both of those make sense - will address!

fwkoch:
- added `SUBMISSION_VERSION_TAG_CHANGE` activity (similar to existing `SUBMISSION_VERSION_STATUS_CHANGE`, etc).
- Renamed `supersedes_submission_version_id` to `supersedes`.
- Removed `superseded_by_submission_version` (this can be retrieved by querying the other `supersedes` version) and `superseded_at` (this can be retrieved from `date` fields)

PR #958, opened 2026-06-16
🧑‍🎓 Unpublish on etl re-register

fwkoch:


stevejpurves:
@fwkoch this is fine as the ETL endpoint here is aimed at works on external storage i.e. that don't need to follow our existing copy-to-pub/rm-from-pub on PUBLISH/UNPUBLISH

PR #1069, opened 2026-08-28
✨ Add tags on submissions (CN-2451)

agutierrezgit:
## Summary

Site admins can classify a submission with **editorial tags** (e.g. “Blog Post”, “Editors Pick”). Kind and collection stay single-valued workflow fields; tags are a separate, multi-valued, site-scoped catalog.

This is phase 1 of [CN-2415](https://linear.app/curvenote/issue/CN-2415) — [CN-2451](https://linear.app/curvenote/issue/CN-2451). Filtering, catalog CRUD, public theme UI, and CLI flags are out of scope.

These are **not** version tags (`SiteWorkDTO.tags` / `v1`, `preprint`, …). Version tags are unchanged. Editorial tags are a first-class `Tag` entity and are exposed on the published payload as `submission_tags`.

### Data
- `Tag` (site-scoped, unique `name` per site) + join table `TagsInSubmissions`
- `name` is derived from the human `label` (`toTagName`): lowercase, URL-safe, 3–32 chars. `label` is bounded to 1–32 chars in the picker, the zod schema, and the loader
- Creating a tag with a label that already exists **reuses** the catalog row (idempotent)
- Removing the last assignment leaves the tag in the catalog (orphans stay — catalog CRUD is the next slice)
- Mutations write `ActivityType.SUBMISSION_TAGS_CHANGE` and bump `Submission.date_modified`, mirroring the kind and date-published actions

### Admin UI (`ee/sites`)
- **Details page:** Tags row between Kind and Slug. One popover to assign, unassign, and create-in-place (Linear-style combobox). Every chip opens it in the *same* place, anchored to the tag row; only `+ Add Tags` anchors to itself. Chips have no remove `x`; unassign from the catalog list. Empty state is `+ Add Tags`, never “Not assigned”.
- **Listing:** display-only row under the dates (up to 3 chips + overflow). No picker, no add/remove.
- **Chips:** solid neutral grey, not outlined, so they do not read as collection / kind / DOI status badges. This adds a `neutral` variant to the shared `ui.Badge` (`packages/scms-core/src/components/ui/badge.tsx`) and a card for it on `system.design`. No existing variant changed.
- Requires `site:submissions:update`. No new scope family. No new v1 write endpoints — create/assign/remove go through the details-page action.

### API (read-only)
- `GET /v1/sites/:siteName` — site tag catalog on `SiteDTO.tags` (also on `GET /v1/sites` and `GET /v1/my/sites`)
- `GET /v1/sites/:siteName/works/:workIdOrSlug/published` — `submission_tags` on that published payload only.
  Each entry is `{ name, label }`: `name` is unique per site and URL-safe, so it is the key external consumers
  filter on. The catalog `id` is an internal write key and is deliberately not exposed here (`TagRefDTO`).
  `SiteDTO.tags` still carries `id` — say the word if you want that narrowed too
- DOI / listing / `SiteWorkDTO.tags` are untouched

### Idempotency contract

`assignTagToSubmission` swallows P2002 and `removeTagFromSubmission` tolerates a zero-row delete, both deliberately. Neither writes a second activity. Both return `TagChangeResult { tag, changed }` so callers can tell a real change from a redundant call — the details-page action fires `SUBMISSION_TAGS_CHANGED` only when `changed` is true, keeping the analytics stream and the activity log in step. The action still returns 200 with the tag either way; the UI is the same in both cases.

### Where to start reviewing
1. Schema + migration: `prisma/schema/submission.prisma`, `prisma/schema/migrations/**/add_submission_tags`
2. Server helpers: `packages/scms-server/src/backend/loaders/sites/tags/`
3. Published payload wiring: `packages/scms-server/src/backend/loaders/sites/submissions/published/`
4. UI: `ee/sites/.../SubmissionTags.tsx`, `TagPicker.tsx`, `SubmissionListingTags.tsx`

## Test plan
- [ ] Details page, with update permission and no tags: enabled `+ Add Tags` opens the picker
- [ ] Type a new label → `Create "…"` assigns it; chips appear; timeline shows the change
- [ ] Same label again reuses the catalog tag (no duplicate)
- [ ] Click a checked row in the picker to unassign; tag remains in the catalog
- [ ] Listing shows assigned tags in their own row; a submission with no tags has no tags row
- [ ] Without update permission: empty state is disabled `Add Tags`; assigned chips are read-only (no `+`)
- [ ] `GET /v1/sites/:siteName` returns the catalog; published work returns `submission_tags`; `tags` on that payload is still version tags
- [ ] DOI response has no `submission_tags`
- [ ] The picker input stops at 32 characters; a longer label posted directly returns a 400 whose reason reaches the UI

Covered in CI: unit specs (name/label validation, picker/listing helpers, `formatSiteDTO`, published DTO), plus `platform/scms/tests/integration/workflow/submission-tags.spec.ts` — including redundant assign/remove reporting `changed: false` with no second activity, the `date_modified` bump, and label-length rejection.

### Verified locally
`bun run typecheck` 24/24 · tag integration suite 15/15, `site-doi-resolve` 19/19 · unit `scms-core` 378, `scms-server` 338, `ee/sites` 111, `common` 2 · prettier + eslint clean (12 pre-existing warnings in unrelated files). The full integration suite has 6 pre-existing failures (`transitions`, `submissions-index-search`, `works.$workId.versions`) — identical before and after these changes, and unrelated to tags.

### Review round 1

- **`build.$jobId.tsx`** — dropped the tag include. The DTO built there never leaves the loader (only `name` and `links.html` are read); the client-facing `site` comes from `sites.get`. Comment added at the call site so the `formatSiteDTO` tripwire does not send someone to add it back.
- **Published payload** — `submission_tags` no longer carries the catalog `id`, per the note above.
- **`SubmissionVersion` in `submission.prisma`** — the realignment is `prisma format` output (prettier has no parser for `.prisma`); the block was already stale on `dev` because of `searchProjection SubmissionSearch?`. Left formatted. `object`, `role` and `user` prisma files are still stale — worth its own PR.
- **`my/sites.server.ts`** — kept: `dbListSitesForUser` backs `GET /v1/my/sites`, not the `/app/sites` grid (that one goes through `dbListSiteCards`, which says so in a comment). `SiteDTO.tags` is required, so dropping the include would return `tags: []` rather than fail to compile.

### Screenshots:

Search or Create new Tags:
Captura de pantalla 2026-08-28 a las 14 56 55

Add Tags:
Captura de pantalla 2026-08-28 a las 14 13 54

Listing View:
Captura de pantalla 2026-08-28 a las 14 57 03

### Review round 2 — tag picker anchoring

Anchoring the popover to *whichever chip was clicked* was more state than the behaviour was worth and carried three defects, so it now follows Linear: chips all open the picker in one fixed place, `+` opens it on itself.

- **Two `ui.Popover` roots, neither ever mounting or unmounting an anchor.** The row root is permanently anchored to the chips `div` and has no trigger; the add root keeps `+` as its only `PopoverTrigger` and never has an anchor. Radix reads `hasCustomAnchor` from state written inside `PopoverAnchor`'s `useEffect`, so a topology that changed on click left the opening render positioned against the wrong control. One `mode: 'none' | 'row' | 'plus'` replaces `pickerOpen` + `anchorTagId`.
- **Focus.** Radix's `onCloseAutoFocus` always returns focus to `triggerRef`, and the row root has no trigger. It is overridden to focus whatever opened the picker, falling back to `+` when that chip was unassigned from the open popover or the control is disabled mid-fetch — otherwise focus silently drops to `<body>`. Chips carry `aria-haspopup="dialog"`; deliberately **not** `aria-expanded`, since they all open the same popover and would each report themselves expanded.
- **Search query.** `PopoverContent`'s exit animation keeps the command mounted ~150 ms, so a quick close/reopen kept the typed query. A session counter, bumped only on a closed → open transition, is now the `key` on `TagPicker`.
- Both roots are non-modal, so each content mounts its own `DismissableLayer`: clicking `+` with the row picker open closes one and opens the other in a single batch. Each root closes with a functional update guarded on its own mode, so the order of the two updates cannot decide the outcome.

Hand-checked in the browser: same position from any chip (including a wrapped second line), `+` under itself, no first-frame jump, no flicker on chip → chip, query reset on reopen, focus back on the opening chip and on `+` after the chip is unassigned, and the read-only / empty states unchanged.


ee/sites/vitest.config.mjs
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
```
agutierrezgit, line 3: It was needed to run the test suite. Apparently the tests here never run

prisma/schema/migrations/20260827153132_add_submission_tags/migration.sql
```diff
@@ -0,0 +1,41 @@
+-- AlterEnum
+ALTER TYPE "ActivityType" ADD VALUE 'SUBMISSION_TAGS_CHANGE';
+
+-- CreateTable
+CREATE TABLE "Tag" (
+    "id" TEXT NOT NULL,
+    "name" TEXT NOT NULL,
+    "label" TEXT NOT NULL,
+    "date_created" TEXT NOT NULL,
+    "site_id" TEXT NOT NULL,
```
stevejpurves, line 10: 👍🏼 scoped to the site

prisma/schema/submission.prisma
```diff
@@ -79,24 +79,24 @@ model SubmissionForm {
...
   submission_activity Activity[]
-  job                 Job?        @relation(fields: [job_id], references: [id])
-  job_id              String?     @unique
+  job                 Job?              @relation(fields: [job_id], references: [id])
+  job_id              String?           @unique
   metadata            Json? // JSON metadata field for extensible data
-  tags                String[]    @default([])
-  occ                 Int         @default(0) // Optimistic concurrency control
+  tags                String[]          @default([])
+  occ                 Int               @default(0) // Optimistic concurrency control
```
stevejpurves, line 99: what did the formatting here? prisma:format or prettier?

prisma/schema/migrations/20260827153132_add_submission_tags/migration.sql
```diff
@@ -0,0 +1,41 @@
...
+CREATE UNIQUE INDEX "Tag_name_site_id_key" ON "Tag"("name", "site_id");
+
+-- CreateIndex
+CREATE INDEX "TagsInSubmissions_tag_id_idx" ON "TagsInSubmissions"("tag_id");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "TagsInSubmissions_submission_id_tag_id_key" ON "TagsInSubmissions"("submission_id", "tag_id");
+
+-- AddForeignKey
+ALTER TABLE "Tag" ADD CONSTRAINT "Tag_site_id_fkey" FOREIGN KEY ("site_id") REFERENCES "Site"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
```
stevejpurves, line 35: ok on it being hard to delete a site 👍🏼

platform/scms/app/routes/build.$jobId.tsx
```diff
@@ -51,7 +51,14 @@ async function dbGetSubmissionVersion(submissionVersionId: string) {
       submission: {
         include: {
           kind: true,
-          site: { include: { submissionKinds: true, collections: true, domains: true } },
+          site: {
+            include: {
+              submissionKinds: true,
+              collections: true,
+              domains: true,
+              tags: { orderBy: { label: 'asc' } },
```
stevejpurves: I don't think we need tags on this API, yet at any rate?

packages/scms-server/src/backend/loaders/my/sites.server.ts
```diff
@@ -60,6 +61,7 @@ async function dbListSitesForUser(user: MyUserDBO) {
             submissionKinds: true,
             collections: { orderBy: { date_created: 'desc' } },
             domains: true,
+            tags: { orderBy: { label: 'asc' } },
```
stevejpurves, line 64: I think this is only used in the main /app/sites listing where we do not show the tags. Could this be omitted then? or are nested routes relying on this data in the parent?

packages/scms-server/src/backend/loaders/sites/submissions/published/get.server.ts
```diff
@@ -178,11 +184,17 @@ export function formatSiteWorkDTO(
...
+export function formatPublishedSubmissionTags(row: PublishedSiteWorkWithTagsRow): TagDTO[] {
+  return row.submission.tags.map((join) => formatTagDTO(join.tag));
+}
+
 export default async function (
   ctx: SiteContext,
   workIdOrSlug: string,
-): Promise<PublishedSiteWorkDTO | null> {
-  const dbo = await dbGetPublishedSiteWorkDto(ctx.site.id, workIdOrSlug);
+): Promise<PublishedSiteWorkWithTagsDTO | null> {
```
stevejpurves, line 195: type extensions are ok here for now, I think we are in this pattern because we're not consistently using or extendin SiteWork as we need it in @curvenote/common, which we'll make easier in our move to monorepo

packages/scms-server/src/backend/loaders/sites/submissions/published/select.server.ts
```diff
@@ -0,0 +1,26 @@
...
+ * shared selects stay narrow so the DOI and listing endpoints keep their
+ * current payload and query cost.
+ */
+export const publishedSiteWorkWithTagsSelect = {
+  ...siteWorkDtoSelect,
+  submission: {
+    select: {
+      ...siteWorkSubmissionSelect,
+      tags: {
+        select: { tag: { select: { id: true, name: true, label: true } } },
```
stevejpurves: should consumers be exposed to `id`? 🤔 if `name` is stable and url-safe then that would be how external consumers interact and how filtering would be acheived?

platform/scms/tests/integration/workflow/submission-tags.spec.ts
```diff
@@ -0,0 +1,360 @@
...
+      },
+    });
+
+    await prisma.tagsInSubmissions.create({
+      data: {
+        id: uuidv7(),
+        date_created: now,
+        tag: { connect: { id: tag.id } },
+        submission: { connect: { id: testData.submissionId } },
+      },
```
stevejpurves, line 36: if we return from a create or update, and don't use the result, we should just `select: { id: true }` otherwise prisma selects whole blob, making unwanted bytes back from the DB. Sometimes that's minimal, sometimes that's KBs, so we've started being explicit

stevejpurves: Ha just realizing that this is in a test! so fine here -- but that rule applies to our app code

stevejpurves (changes requested):
looks good bit a couple of questions there to check @agutierrezgit

prisma/schema/submission.prisma
agutierrezgit: `prisma:format` It seems prettier do not apply to prisma schemas. There are other schemas that need to be aligned but we can do that in other PR  with a CI

platform/scms/app/routes/build.$jobId.tsx
agutierrezgit: Good catch. I'll drop it

packages/scms-server/src/backend/loaders/sites/submissions/published/get.server.ts
agutierrezgit: Yes, I questioned it too but it looks we need it at least for now. I need to understand this a bit better though so I hope to revisit this soon

packages/scms-server/src/backend/loaders/sites/submissions/published/select.server.ts
agutierrezgit: Exactly, not needed. I'll drop it

platform/scms/tests/integration/workflow/submission-tags.spec.ts
agutierrezgit: Yes the app code apply select as expected

prisma/schema/submission.prisma
stevejpurves: yes let's just keep learning where our formatting noise is and we'll deal with that in one go

packages/scms-server/src/backend/loaders/my/sites.server.ts
agutierrezgit: This is the full answer from the agent, let me know if this make sense or there anything we can improve:

This one isn't the /app/sites grid — that listing goes through dbListSiteCards in ee/sites/src/routes/sites/db.server.ts, which has a comment saying it deliberately does not use my.sites / formatSiteDTO. dbListSitesForUser backs GET /v1/my/sites (platform/scms/app/routes/api/v1.my.sites.tsx), so it's the v1 API contract rather than admin UI.

Since SiteDTO.tags is a required field, dropping the include wouldn't fail to compile — formatSiteDTO guards with dbo.tags ?? [], so the endpoint would just return tags: [] for every site, which is wrong rather than absent. That's why both branches of the function have it. Cost is one join per site over three short columns.

On the nested-routes question: nothing depends on it from a parent. The tag picker's catalog is loaded directly by the submission detail loader via sites.tags.dbListSiteTags(ctx.site.id).

prisma/schema/submission.prisma
agutierrezgit: Noted

packages/scms-server/src/backend/loaders/my/sites.server.ts
stevejpurves: hmmm on the CLI is probably the main consumer then. OK leave it on for now, it's not a big payload and will be useful on the CLI

stevejpurves (approved):

Issue #195, opened 2022-06-17
Image Widths & Alignment


rowanc1:
- [x] Schema exports images widths/alignment https://github.com/curvenote/schema/issues/70
- [x] Ensure that single block images also export their markdown correctly
- [x] The curvespace renderer I think should do a calculation

* max: 100%
* % is actually the percent of the ~800px in curvenote
* Set the render in px, not %

I would say test the above in a bunch of situations, and maybe we need:
- [ ] calculate and store the true image width

I am wondering if the badges (which are images) in a lot of repos won't fit this, and we actually need px widths as well?

As part of this, we should also wrap in #180.

PR #1074, opened 2026-08-31
Align agent skills and AGENTS.md

agutierrezgit:
## Summary
- Canonical skills live in `.agents/skills/`; `agent-team` moved there and `.cursor/skills/` is no longer tracked.
- Added `AGENTS.md` (lint + worktrees) and `bun run agents:setup claude` for local Claude Code symlinks. Cursor/Codex need no setup.
- Deleted always-apply Cursor rules that would duplicate `AGENTS.md`.

Closes CN-2462 (curvenote slice).

## Follow-up
The Worktrees section in `AGENTS.md` is the old always-on Cursor rule, copied verbatim. Creating worktrees is a procedure (`bun run wt:create`), so it belongs in a skill (loaded on demand), not in always-on agent memory. Out of scope here; extract in a follow-up.

## Test plan
- [ ] `bun run agents:setup:test`
- [ ] `bun run agents:setup claude` creates `.claude/skills` and `.claude/CLAUDE.md`; second run is a no-op
- [ ] Cursor still sees skills via `.agents/skills/` without running setup
- [ ] `git status` does not show `.claude/CLAUDE.md` / `.cursor/skills` / `.claude/skills`


agutierrezgit:
## For the team

Agent skills live in one place now: `.agents/skills/`. There’s a committed `AGENTS.md` with the always-on norms (lint/format, worktrees). Cursor and Codex read that as-is. Claude Code does not, so it needs a one-time local link.

**Related PRs:** curvenote #1074 · next-theme #195 · dx #1 (clone hint only)

### After this lands (or on this branch)

1. Pull.
2. **Cursor / Codex:** nothing else. Skills and `AGENTS.md` just work. You can skip the script.
3. **Claude Code**, in `curvenote` and in `next-theme`:

```bash
bun run agents:setup claude
```

That creates local (gitignored) links:

- `.claude/skills` → `.agents/skills`
- `.claude/CLAUDE.md` → `AGENTS.md`

Run it twice; the second time is a no-op.

### Local cleanup

Only if setup errors, or you still see the old layout.

- Do **not** delete the whole `.claude/` folder. Keep `settings.json` / `settings.local.json` / worktrees.
- If `agents:setup` says the destination exists and is not a symlink: rename or remove **only** `.claude/skills` (when it’s a real directory, not the new link), then re-run the command.
- Safe to delete leftover **root** `CLAUDE.md` if it’s a symlink to `AGENTS.md` (setup also removes that itself).
- Safe to delete leftover **`.cursor/skills`** if it was the old per-skill symlink farm. Cursor doesn’t use it anymore; skills are in `.agents/skills/`.

We did **not** delete Prisma/Vercel/etc. Those trees were already under `.agents/skills/`. `.cursor/skills/<name>` were just pointers; we stopped tracking the pointers.

New clones: `dx` setup prints the Claude command. Cursor-only folks can ignore it.

PR #1083, opened 2026-09-16
🔧 Fix root scripts that silently skip their bun command

agutierrezgit:
## Problem

`bun --cwd <dir> run <script>` does not run the script. Bun prints the `bun run` usage and exits with code 0, so the calling script looks like it passed.

Reproduced with a minimal package (`"hello": "echo HELLO_RAN"`) on both bun 1.3.10 (the version in `packageManager`) and 1.4.0:

| Command | Result |
| --- | --- |
| `bun --cwd sub run hello` | usage text, exit 0, script not run |
| `bun run --cwd sub hello` | `HELLO_RAN` |
| `bun --cwd sub install` | `error: Script not found "install"`, exit 1 |
| `bun install --cwd sub` | installs, exit 0 |

Affected root commands:

- `wt:go`: the final `check-types` step never ran.
- `test:scms`: the SCMS unit tests never ran.
- `.cursor/worktrees.json`: the last two setup commands failed or did nothing.

## Fix

Move `--cwd` after the subcommand: `bun run --cwd <dir> <script>` and `bun install --cwd <dir>`. There are no other `bun --cwd` uses in the repo.

## Stale test fixed

Once `test:scms` actually ran, one test failed: `app/routes/app/works.$workId.versions/db.server.spec.ts`. `dbLoadWorkVersionsTimeline` returns `versionNumber` since #1010, but the spec's expected objects didn't include it. It went unnoticed because nothing ran the suite; CI doesn't call the SCMS unit tests either. The second commit adds `versionNumber: 2` / `versionNumber: 1` to the expectations. No app code changes.

## Verification

- `bun run --cwd platform/scms check-types` runs and passes.
- `bun run test:scms` runs the suite: 17 files, 105 tests passing.
- `bunx prettier --check` passes on the changed files.


agutierrezgit:
@stevejpurves 

Noticed `wt:go` wasn't actually running the typecheck. `bun --cwd x run y` just prints the bun help and exits 0, so it looked fine. Same thing with `test:scms`, and the setup in `.cursor/worktrees.json`. The fix is moving `--cwd` after `run`.

Once `test:scms` really ran, one test failed. The versions timeline spec didn't expect versionNumber (added in #1010), so I updated it in the same PR. Everything passes now

PR #1086, opened 2026-09-17
⚡️ Make submission tag changes optimistic (CN-2520)

agutierrezgit:
Closes CN-2520.

## Problem

Tag changes on the Submission Details page felt slow once deployed. The picker used one `useFetcher` and locked itself (`pickerBusy`) until the action **and** the full page loader revalidation finished. So every click waited for the whole round trip, and you could not change a second tag in the meantime.

## Change

- **Optimistic:** assign, remove and create show up immediately. What you see is the loader data plus the in-flight submissions, derived in render by pure functions (`pendingTagChanges.ts`, with specs).
- **Several tags at once:** each tag name gets its own keyed fetcher (`useSubmit` with `navigate: false` and `fetcherKey`), and pending changes are read with `useFetchers()`. Only the tag with a change in flight is blocked; its picker row is disabled until it settles. A newly created tag keys on its `name`, which the client derives with the same `toTagName` the server uses.
- **Errors:** a failed change reverts on its own when its fetcher goes idle, and a toast (`ui.toastError`) replaces the inline red text. `useFetchers()` does not expose fetcher data, so `TagChangeErrorToast` watches each tag fetcher by key.
- **Reload after a failed change:** the route's `shouldRevalidate` reloads the page data after tag actions even when they fail. React Router skips revalidation after a 4xx/5xx action by default, which left a tag deleted in another tab on screen, and every retry failed the same way.
- `tag-assign` / `tag-remove` are added to `SUBMISSION_DETAIL_FORM_ACTIONS`.
- No changes to the action or the tag services.

### Why not the one-fetcher pattern used by the other fields

Slugs, kind and collection use one `useFetcher` per control and read `fetcher.formData`. One fetcher holds only one pending change: a second submit replaces the first `formData`, so the first chip would disappear until revalidation. The hook has a comment explaining this.

### Accepted trade-offs

- The optimistic list sorts with `localeCompare`; Postgres collation may place a few labels differently, so a chip can shift position when server data arrives.
- Every tag touched during a page visit leaves a fetcher registered until the user leaves the page (at most one per catalog tag).

## Manual test (Slow 3G in devtools)

| # | Action | Expected |
|---|---|---|
| 1 | Open picker, select tag A then tag B quickly | Both chips and check marks appear immediately; A and B rows disabled until each settles |
| 2 | While B is in flight, deselect assigned tag C | C disappears immediately; B unaffected |
| 3 | Reload after all settle | Server state matches what was shown |
| 4 | Type a new label in the picker, pick `Create "…"`, keep the picker open | Chip appears immediately; in the same dropdown the `Create` row becomes a disabled, checked row for the new tag until created |
| 5 | Click a busy row again | No extra POST |
| 6 | Delete tag D in another tab, then toggle D | Toast `tag not found on this site`; D disappears from the chips and the picker once the page data reloads |
| 7 | Change slug / kind / collection | Tags unaffected, no tag toast |
| 8 | Close and reopen the picker mid-flight | Pending state and disabled row persist |

- [x] Manual test passed (all eight scenarios, Slow 3G)

Also from the manual round: the picker's search box now clears after each pick, so tags can be chained without clearing it by hand.


ee/sites/src/routes/$siteName.submissions.$submissionId/pendingTagChanges.ts
```diff
@@ -0,0 +1,119 @@
...
+  { kind: 'assign' | 'remove'; name: string } | { kind: 'create'; name: string; label: string };
+
+/** The fields of a `useFetchers()` entry read here, so specs do not need a router. */
+export type TagFetcherSnapshot = {
+  key: string;
+  state: 'idle' | 'loading' | 'submitting';
+  formData?: FormData;
+};
+
+function keyPrefix(submissionId: string): string {
```
stevejpurves, line 25: @agutierrezgit keyed fetchers are fine, work well and are react-router native but a simpler pattern that I prefer is to wrap the controls that are repeated e.g. in this case the `tag` in a tiny react component with it's own `useFetcher` inside of that component. RR then automatically managed the fetcher per component without you having to explicitly manage a key.

If you are not needing to access that key and only using keys to avoid sharing state across a bunch on actions then that can be a better pattern.

Although here I see you the use of the Toasts probably calls for the explicit key pattern - so this is ok

stevejpurves (approved):
This is working well in local testing on 4G (and below!) network throttle.

PR #1084, opened 2026-09-16
✨ Site DOI foundation: settings, scopes and Crossref lookup client (CN-2507)

agutierrezgit:
Linear: [CN-2507](https://linear.app/curvenote/issue/CN-2507) · Part of [CN-2501](https://linear.app/curvenote/issue/CN-2501) (DOI registration MVP)

## What this does

First step of self-serve DOI registration. It adds what a site needs to hold its DOI setup. There is no UI yet and nothing talks to Crossref in production.

- **Site DOI settings.** A new `SiteDoiConfig` table, one row per site. Each row holds:
  - whose prefix the site uses (`CURVENOTE_PREFIX` or `CUSTOM_PREFIX`);
  - the prefix and the Crossref role;
  - a status (`PENDING_ROLE`, `ACTIVE`, `NEEDS_ATTENTION`).

  A custom prefix and role can belong to only one site. Sites on Curvenote's prefix all share it. One migration, no data changes.
- **Permissions.** New `site:doi` scopes. Site admins can read, register and configure. Members can only read.
- **Crossref config.** Host, depositor email and password per deployment: `test.crossref.org` outside production, `doi.crossref.org` in production. A missing or malformed value fails with the field name, never the value.
- **Crossref client** (in `ee/sites`). Two calls, covered by tests against real responses we recorded from Crossref's test system: look up who owns a prefix, and check that a role can log in.

## Where to look

1. `prisma/schema/doi.prisma` and its migration. The uniqueness rule is a partial index written in raw SQL, because Prisma cannot express it.
2. `packages/scms-server/src/backend/roles.server.ts`: who gets which DOI scope
3. `ee/sites/src/backend/crossref/client.server.ts`

## Next

**[CN-2508](https://linear.app/curvenote/issue/CN-2508): site DOI setup.** This builds on this PR.
- The Site > DOI Registration screen: choose Curvenote's prefix or the site's own.
- A system admin binds the Crossref role.
- Validation uses this client and moves the site to `ACTIVE`.

Items carried over from this PR:
- Normalise the prefix before the lookup.
- Send a `User-Agent` with a contact email to api.crossref.org.
- On a 401, tell a bad role apart from a bad system password.
- Reject Curvenote's own prefix on a `CUSTOM_PREFIX` site.
- Validate action input with zod.

[...]

PR #1087, opened 2026-09-17
✨ Site DOI setup: registration method, role binding and Crossref validation (CN-2508)

agutierrezgit:
Linear: [CN-2508](https://linear.app/curvenote/issue/CN-2508) · Part of [CN-2501](https://linear.app/curvenote/issue/CN-2501) (DOI registration MVP)

> **Draft. Stacked on #1084 (CN-2507)**, so the diff here is CN-2508 only. Retarget to `dev` once #1084 merges.
> The UI rework is in: the Enterprise dialog is gone (the locked option and the upsell are inline now), the setup screen is split into smaller cards, `normalizePrefix` accepts a bare registrant code (`5555`), and setup and reset no longer return a success message. Linking the role still has its own card, not the Registration account card.

## TL;DR

Site > DOI Registration. A site admin sets up how the Site registers DOIs. With the Enterprise flag on, a Curvenote system admin links the Site's Crossref role. The role is validated against Crossref before the Site becomes `ACTIVE`. Nothing is deposited yet; that is phase 2 ([CN-2518](https://linear.app/curvenote/issue/CN-2518)).

## Screens

Captured from a local development database, in the order a Site walks through them. Each step
shows the screen in place, then the card itself where the detail matters.

### 1. First-time setup, without the Enterprise flag

The default for every Site. Own prefix is shown locked next to the included option, with the
upsell inline — the old dialog is gone.

Set up DOI registration

The registration method picker

### 2. Curvenote-managed, once configured

`ACTIVE` at once, no Crossref call. The prefix is locked and the card names Curvenote as the
owner, with the Enterprise upsell underneath.

DOI registration status: Active

Registration account, Curvenote-managed

### 3. The Enterprise flag, Site > Advanced

System admin only. It gates the own-prefix option; turning it off does not change a Site that is
already configured.

Allow own DOI prefix (Enterprise)

### 4. First-time setup, with the flag on

Straight to the prefix form. The field accepts `doi:10.5555`, `https://doi.org/10.5555/` and the
bare registrant code `5555`.

Connect your organization's Crossref account

### 5. Waiting for the Crossref role

After the prefix resolves at Crossref. The Site stays in `PENDING_ROLE`, the prefix is still
editable, and the alert tells the site admin exactly what to do in their Crossref account.

DOI registration status: Waiting for Crossref role

Registration account, pending the role

### 6. Linking the role, system admin only

The card repeats the prefix owner so the admin can check it against the Crossref email before
linking. Nothing is saved unless the Crossref login passes.

Link the Crossref role

### 7. Advanced actions and the confirmation

Reset names the Site and says what is lost. Unlink joins it here once a role is linked.

Advanced actions

Reset DOI setup?

## What this does

- **Two ways to register.**
  - **Curvenote-managed:** Curvenote's prefix and role come from config. The Site is `ACTIVE` at once and Crossref is not called.
  - **Own prefix (Enterprise):** the site admin enters the prefix. We normalise it (`doi:10.5555`, `https://doi.org/10.5555/`), reject Curvenote's own prefix, and look up its owner at Crossref. The Site waits in `PENDING_ROLE` until a system admin links the role.
- **Linking the role** is the one manual step. The system admin enters the role from the Crossref email. We refresh the prefix owner, then log in to Crossref as `<depositorEmail>/<role>`.
  - On a 401 we log in once more with Curvenote's own role. If that also fails, the system password is wrong, not the role.
  - A prefix and role pair already held by another Site is refused.
- **System admin actions:** link the role and unlink it (back to `PENDING_ROLE`). Reset (delete the setup) is the site admin's too until a role is linked, because it is the only way to undo their own choice of method or prefix; after that it is system-admin only. Unlink and reset have a hook for "no registered DOIs yet". It is a no-op until CN-2518 adds the registration model.
- **Enterprise flag:** Site > Advanced gets a system-admin-only checkbox (`site.data.doiCustomPrefixEnabled`). Turning it off does not change a Site that is already configured.
- **Every write is one transaction** with a `SITE_DOI_CONFIG_UPDATED` activity (who, what, row after the write). "Linked by X on date" is read from that log.
- **Stale tabs:** every form posts `occ`. A write that lost the race gets "Configuration changed, reload the page" and never overwrites.
- **Config:** `api.crossref` gains `prefix` and `role` (Curvenote's own). Both are required. Prefix lookups now send a `User-Agent` with the depositor email, as api.crossref.org asks.

## Who can do what

| Action | Member | Site admin | System admin |
| --- | --- | --- | --- |
| See Site > DOI Registration | ❌ | ✅ | ✅ |
| Set up Curvenote-managed | ❌ | ✅ | ✅ |
| Set up / correct own prefix (flag on, before the role is linked) | ❌ | ✅ | ✅ |
| Link / unlink the Crossref role | ❌ | ❌ | ✅ |
| Reset the setup, before Curvenote links a role | ❌ | ✅ | ✅ |
| Reset the setup, once a role is linked | ❌ | ❌ | ✅ |
| Toggle the Enterprise flag | ❌ | ❌ | ✅ |

The system-admin checks run in the services and in the Advanced action, not only in the UI.

## Where to look

1. `ee/sites/src/backend/doi/role.server.ts` and `configure.server.ts`: the six intents, their guards, and what each one writes.
2. `ee/sites/src/backend/doi/validation.server.ts`: prefix resolution and the two-login role check.
3. `ee/sites/src/backend/doi/db.server.ts`: `commitDoiWrite` maps P2002 and P2025 to results.
4. `ee/sites/src/routes/$siteName.doi/actionHelper.server.ts`: form parsing and intent dispatch.
5. UI in `ee/sites/src/routes/$siteName.doi/`. Skim it for now, it is being reworked.

The commits are in dependency order (normaliser → config → service → intents → flag → route → UI) and can be read one by one.

## Tests

Unit tests with Prisma and `fetch` stubbed: every intent's success path, each guard, and each Crossref failure, with nothing written on failure. On the pushed commits, `bun run --filter @curvenote/scms-sites-ext test` passes 320 tests over 30 files, with `compile` and the Prettier check clean and no new lint warnings.

Not done yet: a manual run against `test.crossref.org` with a real role.

## Deploy notes

`api.crossref.prefix` and `api.crossref.role` must be set in each deployment before this ships. Without them the screen shows "DOI registration is not configured" and the action returns 500.


ee/sites/src/backend/crossref/prefix.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Reduce what a person pastes ("doi:10.5555", "https://doi.org/10.5555/") or types ("5555") to the
+ * bare prefix. Returns null when the result is not a DOI prefix, so callers show one form error.
+ */
+export function normalizePrefix(raw: string): string | null {
+  const stripped = raw.trim().replace(LEADING, '').trim().replace(/\/+$/, '');
+  const prefix = CODE_ONLY.test(stripped) ? `10.${stripped}` : stripped;
+  return PREFIX_RE.test(prefix) ? prefix : null;
+}
```
stevejpurves: @agutierrezgit we have a solid dependency `doi-utils` that we use for normalisation and matching purposes https://www.npmjs.com/package/doi-utils

lets add an AGENT.md line on this too

ee/sites/src/backend/doi/db.server.ts
```diff
@@ -0,0 +1,171 @@
...
+ * Prisma's schema does not know. P2025 (no row matched `id` + `occ`) means another request changed
+ * or removed the row.
+ */
+export async function commitDoiWrite(
+  deps: DoiDeps,
+  write: DoiWrite,
+  fn: (tx: DoiTx) => Promise<DoiConfigRow | null>,
+): Promise<DoiResult> {
+  try {
+    const row = await deps.prisma.$transaction(async (tx) => {
```
stevejpurves: we don't usually pass prisma around, as we have a singleton via getPrismaClient we just request is where we need it

ee/sites/src/backend/doi/role.server.ts
```diff
@@ -0,0 +1,131 @@
...
+import { ROLE_RE, lookupOwner, validateRole } from './validation.server.js';
+
+const FORBIDDEN: DoiFailure = { ok: false, status: 403, error: DOI_ERRORS.forbidden };
+const PAIR_TAKEN: DoiFailure = { ok: false, status: 409, error: DOI_ERRORS.pairTaken };
+
+export type BindRoleInput = {
+  siteId: string;
+  actor: DoiActor;
+  role: string;
+  occ: number;
```
stevejpurves: I was confused by the occ here, but i see that is exclusively for the new SiteDoi UI and not our shared OCC mechanisms on our `.data` fields.

ee/sites/src/routes/$siteName.advanced/doiFlag.server.ts
```diff
@@ -0,0 +1,44 @@
...
+    const prisma = await getPrismaClient();
+    const existing = await prisma.site.findUnique({
+      where: { id: ctx.site.id },
+      select: { data: true },
+    });
+    const existingData = (existing?.data as Record<string, unknown> | null) ?? {};
+    await prisma.site.update({
+      where: { id: ctx.site.id },
+      data: {
+        data: { ...existingData, doiCustomPrefixEnabled: payload.doiCustomPrefixEnabled },
```
stevejpurves: @agutierrezgit whenever we update a JSONB field we should be using our OCC mechanism, there is a pattern for this in the codebase, look at `WorkVersion.metadata` updates for example. We want to put that in place whenever we modify a json field that there is a change of collision on.

stevejpurves: that is only FYI in this case, as this is a system-admin only flag with very low likelihood of collision then we can skip it.

jillianlhale:
@agutierrezgit this looks really good! I just have one small update for step 6: Could we please change "organisation" to "organization" and "authorisation" to "authorization"?

Since many of our larger customers are based in the US, our product standard should probably default to **American spelling** for any customer-facing copy or instructions (e.g., optimize, color, center). This ensures consistency across the entire user experience. 

If not done already, we could also update the system prompts we give Claude. We should explicitly instruct the AI to always output customer-facing text using American English spelling rules.

cc @stevejpurves

agutierrezgit:
Done! Switched to "organization" / "authorization" and fixed a couple more spots in the DOI screens. Not sure why Claude went British on us 😅 but I've added a note to `AGENTS.md` so it sticks to American spelling from now on.

